PR #27, opened 2021-10-01
Remove NODE_ENV conditional so sequence timeline shows in `production`

### LINKED ISSUE #26: `NODE_ENV` set to production disables sequence timeline and ability to edit keyframes
## What?
When building a project that uses the `studio` npm package - and `NODE_ENV` is set to `production` the sequence pane (timeline) doesn't show up and keyframes become un-editable

## Steps to reproduce
- Create a new project with `create-react-app` and add `@theathre/core` / `@theathre/studio`
- Run `build` command
- Serve the built project, no timeline appears and the properties become un-editable 
- (Animations still play if I hit the space bar with the animated object selected - but sequence editor does not show / keyframes are not editable)

### When `NODE_ENV` is set to `development` this is what shows
dev

### When `NODE_ENV` is set to `production` this is what shows
prd

juniorxsound:
## What?
- [x] Remove `NODE_ENV` conditional guard from `getMapOfValidSequenceTracks_forStudio` method in `SheetObjectTemplate.ts` so sequence timeline shows up when `NODE_ENV` is set to `production` (useful for Create React App's where `NODE_ENV` is controlled by `react-scripts`)

Closes #26 

## Additional context
When building `examples/dom-cra` it now properly shows the timeline/sequence editor and keyframes are editable
fix

### DIFF
--- a/theatre/core/src/sheetObjects/SheetObjectTemplate.ts
+++ b/theatre/core/src/sheetObjects/SheetObjectTemplate.ts
@@ -174,21 +174,17 @@ export default class SheetObjectTemplate {
    * Not available in core.
    */
   getMapOfValidSequenceTracks_forStudio(): IDerivation<IPropPathToTrackIdTree> {
-    if (process.env.NODE_ENV !== 'production') {
-      return this._cache.get('getMapOfValidSequenceTracks_forStudio', () =>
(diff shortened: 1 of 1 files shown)


AriaMinaei:
Thanks for the fix @juniorxsound! This is now published in `0.4.6`.

PR #29, opened 2021-10-01
WIP: CONTRIBUTING

jmyrland:
Todos:
- [x] CONTRIBUTING.md skeleton
- [ ] build packages/react? (required?)
- [ ] build packages/dataverse? (required?)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -49,13 +49,15 @@ git remote add upstream git@github.com:AriaMinaei/theatre.git
 
 ### `@theatre/core`
 
-The core animation library. This project is located under
-[`./theatre/core`](`/theatre/core).
+The core animation library.
+
+This project is located under [`./theatre/core`](./theatre/core).
 
 ### `@theatre/studio`
 
-The visual editor. This project is located under
-[`./theatre/studio`](`/theatre/studio).
+The visual editor.
+
+This project is located under [`./theatre/studio`](./theatre/studio).
 
 ### Commands
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -123,7 +123,8 @@ Other references
 
 ## Testing
 
-Run tests during development with `yarn test` to re-run tests on file changes.
+Run tests during development with `yarn test --watch` to re-run tests on file
+changes.
 
 ### Examples
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,4 +1,134 @@
 # Contributing to Theatre.js
 
->`@todo` Write the contribution guide.
+## Getting Started
 
+### Prerequisites
+
+Make sure you have [Node](https://nodejs.org/) installed at v14.0.0+
+
+```sh
+node -v
+
+v14.0.0
+
+```
+
+and [Yarn](https://classic.yarnpkg.com/en/) at v2.0.0+.
+
+```sh
+yarn -v
+
+2.0.0
+```
+
+### Fork, Clone & Install
+
+Start by forking Theatre.js to your GitHub account. Then clone your fork and
+install dependencies:
+
+```sh
+git clone git@github.com:<your-user>/theatre.git
+cd theatre
+yarn
+```
+
+> ⚠ theatre relies on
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -1,4 +1,134 @@
...
+Make sure you have [Node](https://nodejs.org/) installed at v14.0.0+
+
+```sh
+node -v
+
+v14.0.0
+
+```
+
+and [Yarn](https://classic.yarnpkg.com/en/) at v2.0.0+.
```
jmyrland, line 16: I've got 2.4.2 currently installed, so I just set 2.0.0 as a minimum requirement.

CONTRIBUTING.md
```diff
@@ -1,4 +1,134 @@
...
+## Getting Started
 
+### Prerequisites
+
+Make sure you have [Node](https://nodejs.org/) installed at v14.0.0+
+
+```sh
+node -v
+
+v14.0.0
```
jmyrland, line 12: I've got 14.15.4 currently installed, so I just set 14.0.0 as a minimum requirement.

CONTRIBUTING.md
```diff
@@ -1,4 +1,134 @@
...
+
+1. Run the playground to verify the changes
+   ```
+   yarn playground
+   ```
+2. Watch and rebuild theatre packages on changes:
+   ```sh
+   cd theatre
+   yarn build:js
+   ```
```
jmyrland, line 97: Not entirely sure about this? It seems to be working for me. If I add a console log inside the `@theatre/studio`, it does appear after i refresh the playground 👍 

However the `playground` is just a blank screen with the studio and a couple of console errors.

jmyrland (commented):
I wasn't able to build the `packages/react` or `packages/dataverse` modules 🤔

Also, is the `playground` supposed to just be a blank screen with just  the studio?

CONTRIBUTING.md
jmyrland: image

CONTRIBUTING.md
AriaMinaei: I think with Yarn, you install 1.x.x globally, and when you `cd` into a package that uses yarn `2.x`, it switches to 2.x automatically. Yarn 2.x's whole bundle is even [committed](https://github.com/AriaMinaei/theatre/blob/69c6aa9af226ca4dc852cad1145b9cd0f42f9647/.yarn/releases/yarn-2.4.2.cjs) to the repo.

AriaMinaei: So we could say something like:

> Install the latest version of yarn 1.x globally (`npm install -g yarn`).

CONTRIBUTING.md
AriaMinaei: Ah so playground is made so you don't have to build any of the packages. It has an esbuild config that automatically includes all the packages. So, if you want to get started hacking, you can just `yarn playground` and that should be enough.

AriaMinaei: However, I think the stuff in `packages/playground/src` is a bit confusing. Like, what's supposed to happen with any of those playgrounds directories?

I think the goal of playground should be like this:
1. It's a quick cycle way to hack on theatre. No long builds, no wait time, just a single esbuild setup for all the packages.
2. `packages/playground/src/index.ts` just imports one of the playground directories, like so: `import './shared/audio/index.ts` or something like that. This `playground/src/index.ts` file is actually not committed to the repo. It will be created first time you run `yarn playground` and is `.gitignore`d. This allows you to change its `import` and for example, say `import './local/walk-cycle/index.ts`, without having to commit `src/index.ts` or `src/local` to the repo.
3. We'll have some ready-made playgrounds to in `packages/playground/src/shared/`, which _will_ be comitted to the repo.

AriaMinaei: I'll do that playground thing now, so we can get the PR merged soon.

mattwel:
UNSubscribe

CONTRIBUTING.md
AriaMinaei: Ok, playground is done.

AriaMinaei:
This is merged (though not sure why Github isn't marking it as merged). Thanks @jmyrland!

mattwel:
Unsubscribe

Issue #34, opened 2021-10-08
Inconsistent keyboard shortcut for opening the studio


jmyrland:
Opening the studio requires pressing different keys (`alt + '` on a Norwegian layout), depending on which keyboard layout is used.

There should be a unified way of opening the studio, that is independent of the keyboard layout.

The issue lies here: https://github.com/AriaMinaei/theatre/blob/main/theatre/studio/src/UIRoot/useKeyboardShortcuts.ts#L47

Would replacing `e.code === 'Backslash'` with `e.code === "'"` solve this issue for all layouts?

PR #35, opened 2021-10-09
Fixes inconsisten keyboard shortcuts for opening studio

jmyrland:
Adds support for `alt + \` by key (while still supporting the old shortcut `e.code === 'Backslash'` and adding support for `e.code === 'IntlBackslash'`).

Updates the logged message when the studio is hidden:
> Theatre Studio is hidden. Use the keyboard shortcut 'alt + \\' to restore the studio, or call studio.ui.restore().

With these changes, `alt + \` should in *most cases* toggle the studio visibility.

related to #34

### DIFF
--- a/theatre/studio/src/UIRoot/UIRoot.tsx
+++ b/theatre/studio/src/UIRoot/UIRoot.tsx
@@ -71,7 +71,7 @@ export default function UIRoot() {
   useEffect(() => {
     if (visiblityState === 'everythingIsHidden') {
       console.warn(
-        `Theatre Studio is hidden. You can restore it by either hitting alt/option+\` or calling studio.ui.restore()`,
+        `Theatre Studio is hidden. Use the keyboard shortcut 'alt + \\' to restore the studio, or call studio.ui.restore().`,
       )
     }
     return () => {}
--- a/theatre/studio/src/UIRoot/useKeyboardShortcuts.ts
+++ b/theatre/studio/src/UIRoot/useKeyboardShortcuts.ts
@@ -44,7 +44,12 @@ export default function useKeyboardShortcuts() {
         } else {
           return
         }
-      } else if (e.key === `'` && e.altKey) {
+      }
+      // alt + \
+      else if (
+        e.altKey &&
+        (e.key === '\\' || e.code === 'Backslash' || e.code === 'IntlBackslash')
+      ) {
         studio.transaction(({stateEditors, drafts}) => {
           stateEditors.studio.ahistoric.setVisibilityState(
             drafts.ahistoric.visibilityState === 'everythingIsHidden'


AriaMinaei:
The problem with using `e.key` is that, depending on the keyboard layout of the user, the physical location of the key would change a lot. The `'` key for example moves all over the place on different layouts:

grafik
grafik
grafik
grafik

`e.code` OTOH is supposed to remain constant in terms of the physical location of the keys. From the [MDN article](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code):

> The `KeyboardEvent.code` property represents a physical key on the keyboard (as opposed to the character generated by pressing the key). In other words, this property returns a value that isn't altered by keyboard layout or the state of the modifier keys. 

Our issue here is the `Backslash` key, which [others](https://github.com/microsoft/vscode/issues/24153) seem to have experienced as well.

From my late night search, the `Backslash` key is apparently special in that it exists on some physical layouts, and in others, it is replaced by `IntlBackslash`. From W3C's [UI Event codes](https://www.w3.org/TR/uievents-code/#keyboard-102):

> The standard "102" keyboard is common throughout Europe and adds a key that doesn’t exist on the "101" layouts: The "IntlBackslash" key (labelled \| on a UK keyboard) next to the left shift key.
>
> A second key is also added (labelled #~ on a UK keyboard) which is partially tucked under the Enter key. This key is encoded as "Backslash", using the same code as the \| key found on the "101" keyboard layout. According to [USB-HID], the US \| and UK #~ are actually two separate keys (named "Keyboard \ and |" and "Keyboard Non-US # and ~"), but since these two keys never co-occur on the same keyboard most platforms use the same scancode for both keys, making them difficult to distinguish. It is for this reason that the code "Backslash" is used for both of these keys.
> grafik


So, if I'm not mistaken, the best way to handle this is to stick with `e.code` check for either `Backslash` or `IntlBackslash` to trigger the event.

jmyrland:
> The problem with using `e.key` is that, depending on the keyboard layout of the user, the physical location of the key would change a lot. The `'` key for example moves all over the place on different layouts:
> 
> grafik
> grafik
> grafik
> grafik
> 
> `e.code` OTOH is supposed to remain constant in terms of the physical location of the keys. From the [MDN article](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code):
> 
> > The `KeyboardEvent.code` property represents a physical key on the keyboard (as opposed to the character generated by pressing the key). In other words, this property returns a value that isn't altered by keyboard layout or the state of the modifier keys. 
> 
> Our issue here is the `Backslash` key, which [others](https://github.com/microsoft/vscode/issues/24153) seem to have experienced as well.
> 
> From my late night search, the `Backslash` key is apparently special in that it exists on some physical layouts, and in others, it is replaced by `IntlBackslash`. From W3C's [UI Event codes](https://www.w3.org/TR/uievents-code/#keyboard-102):
> 
> > The standard "102" keyboard is common throughout Europe and adds a key that doesn’t exist on the "101" layouts: The "IntlBackslash" key (labelled \| on a UK keyboard) next to the left shift key.
> >
> > A second key is also added (labelled #~ on a UK keyboard) which is partially tucked under the Enter key. This key is encoded as "Backslash", using the same code as the \| key found on the "101" keyboard layout. According to [USB-HID], the US \| and UK #~ are actually two separate keys (named "Keyboard \ and |" and "Keyboard Non-US # and ~"), but since these two keys never co-occur on the same keyboard most platforms use the same scancode for both keys, making them difficult to distinguish. It is for this reason that the code "Backslash" is used for both of these keys.
> > grafik
> 
> 
> So, if I'm not mistaken, the best way to handle this is to stick with `e.code` check for either `Backslash` or `IntlBackslash` to trigger the event.
> 
> 

Whoa that's some proper research! Alright, I'm convinced.

Should I add support for both `Backslash` and `IntlBackslash`, or should we just close this PR and the related issue?

AriaMinaei:
> Should I add support for both Backslash and IntlBackslash

That would be great!

AriaMinaei:
oops, that console log is actually wrong. It should say alt and backslash. And I think it should add a line to explain we mean the physical backslash key, so it might be a different character on different layouts.

I see that many different tools use language-specific keyboard shortcuts (so alt+backslash of English would be alt+apostrophe in some other language).

I don't think we have the bandwidth right now to design a way to communicate different keys for each language, but since theatre's users are all technical right now, we can just explain this in the console log.

I can try this in a couple of days, or feel free to amend the PR and I'll merge it.

jmyrland:
> ... `@` in Norwegian and so forth).

Hehe, maybe - IF you're on a mac 😁 

I've made the changes, and `alt + \` is now supported on my end at least, while still supporting the old shortcut:
nTezaF2uSs

If you're happy with the changes, feel free to merge it whenever 👍

PR #36, opened 2021-10-12
R3F: Add ability to provide additional props to sheet object and get sheet ref

juniorxsound:
## What?
- [x] Add ability to provide additional props to R3F components that use the `editable` extension to be able to add custom props to the `sheet.object`
- [x] Add ability to get the sheet ref from the component by providing a ref to the component

additional_props
> Example of custom `shadows` prop being used on the  `<e.mesh>` element in the playground example

## How to use?
```jsx
function App(){
  const sheetRef = useRef();

  useEffect(() => {
    sheetRef.current.onValuesChange(newValues => {
      console.log(newValues)
    });
  }, [sheetRef ])

  return(
  <e.group additionalProps={{ someCustomProp: 'yay' }} objRef={sheetRef} />
  )
}
```

### DIFF
--- a/packages/r3f/src/components/editable.tsx
+++ b/packages/r3f/src/components/editable.tsx
@@ -85,7 +85,7 @@ const editable = <
         useEditorStore
           .getState()
           .setSheetObject(uniqueName, sheetObject as $FixMe)
-      }, [sheet, uniqueName, additionalProps])
+      }, [sheet, uniqueName])
 
       const transformDeps: string[] = []
 


AriaMinaei:
I added a review comment, but other than that, this PR is good to merge! I think we're getting close to shipping `@theatre/r3f`, thanks to @juniorxsound 🙏

AriaMinaei:
Merged! Thanks @juniorxsound! 🙏

PR #41, opened 2021-10-19
Allow studio to be rendered in specific html element

cory-glooh:


AriaMinaei:
Ah, this PR went stale since we spoke about it on Discord. We decided that rendering in a custom element may open up more surface area for design inconsistencies so we'll hold off on that until more of the design is fleshed out.

PR #55, opened 2021-12-09
Adds codesandbox/ci.json

gsimone:


gsimone:
@AndrewPrifer if you add the codesandbox app, this should work

PR #64, opened 2022-01-25
Create color prop

AndrewPrifer:
Stuff missing:

- [x] Create proper UI
- [x] Add sanitizer function and validate and sanitize default value
- [ ] Add docs
- [x] Add missing gamma conversion to the interpolator that I forgot
- [x] Fix `SerializablePrimitive` type which currently breaks CI. The color prop is the first primitive (non-compound) prop that is not a `string | number | boolean` and apparently just adding `Object` to the `SerializablePrimitive` type didn't do the trick, and doesn't make much sense from a type-safety point of view either.
- [x] Create a color picker that integrates seamlessly with theatre

### DIFF
--- a/theatre/core/src/propTypes/index.ts
+++ b/theatre/core/src/propTypes/index.ts
@@ -1,7 +1,13 @@
 import type {$IntentionalAny} from '@theatre/shared/utils/types'
 import userReadableTypeOfValue from '@theatre/shared/utils/userReadableTypeOfValue'
-import type {Rgba, Laba} from '@theatre/shared/utils/color'
-import {decorateRgba} from '@theatre/shared/utils/color'
+import type {Rgba} from '@theatre/shared/utils/color'
+import {
+  decorateRgba,
+  linearSrgbToOklab,
+  oklabToLinearSrgb,
+  srgbToLinearSrgb,
+  linearSrgbToSrgb,
+} from '@theatre/shared/utils/color'
 import {mapValues} from 'lodash-es'
 import type {
   IShorthandCompoundProps,
@@ -272,73 +278,6 @@ const _interpolateRgba = (
   right: Rgba,
   progression: number,
 ): Rgba => {
-  // TODO: Factor out all these utility functions
-  function linearSrgbToSrgb(rgba: Rgba) {
-    function compress(x: number) {
-      // This looks funky because sRGB uses a linear scale below 0.0031308 in
-      // order to avoid an infinite slope, while trying to approximate gamma 2.2
-      // as closely as possible, hence the branching and the 2.4 exponent.
-      if (x >= 0.0031308) return 1.055 * x ** (1.0 / 2.4) - 0.055
-      else return 12.92 * x
-    }
-    return {
-      r: compress(rgba.r),
-      g: compress(rgba.g),
-      b: compress(rgba.b),
-      a: rgba.a,
-    }
-  }
-
-  function srgbToLinearSrgb(rgba: Rgba) {
-    function expand(x: number) {
-      if (x >= 0.04045) return ((x + 0.055) / (1 + 0.055)) ** 2.4
-      else return x / 12.92
-    }
-    return {
-      r: expand(rgba.r),
-      g: expand(rgba.g),
-      b: expand(rgba.b),
-      a: rgba.a,
-    }
-  }
-
-  function linearSrgbToOklab(rgba: Rgba) {
-    let l =
-      0.4122214708 * rgba.r + 0.5363325363 * rgba.g + 0.0514459929 * rgba.b
-    let m =
-      0.2119034982 * rgba.r + 0.6806995451 * rgba.g + 0.1073969566 * rgba.b
-    let s =
-      0.0883024619 * rgba.r + 0.2817188376 * rgba.g + 0.6299787005 * rgba.b
-
-    let l_ = Math.cbrt(l)
-    let m_ = Math.cbrt(m)
-    let s_ = Math.cbrt(s)
-
-    return {
-      L: 0.2104542553 * l_ + 0.793617785 * m_ - 0.0040720468 * s_,
-      a: 1.9779984951 * l_ - 2.428592205 * m_ + 0.4505937099 * s_,
-      b: 0.0259040371 * l_ + 0.7827717662 * m_ - 0.808675766 * s_,
-      alpha: rgba.a,
-    }
-  }
-
-  function oklabToLinearSrgb(laba: Laba) {
-    let l_ = laba.L + 0.3963377774 * laba.a + 0.2158037573 * laba.b
-    let m_ = laba.L - 0.1055613458 * laba.a - 0.0638541728 * laba.b
-    let s_ = laba.L - 0.0894841775 * laba.a - 1.291485548 * laba.b
-
-    let l = l_ * l_ * l_
-    let m = m_ * m_ * m_
-    let s = s_ * s_ * s_
-
-    return {
-      r: +4.0767416621 * l - 3.3077115913 * m + 0.2309699292 * s,
-      g: -1.2684380046 * l + 2.6097574011 * m - 0.3413193965 * s,
-      b: -0.0041960863 * l - 0.7034186147 * m + 1.707614701 * s,
-      a: laba.alpha,
-    }
-  }
-
   const leftLab = linearSrgbToOklab(srgbToLinearSrgb(left))
   const rightLab = linearSrgbToOklab(srgbToLinearSrgb(right))
 
--- a/theatre/core/src/propTypes/index.ts
+++ b/theatre/core/src/propTypes/index.ts
@@ -273,20 +273,21 @@ export const rgba = (
         `Argument defaultValue in t.rgba(defaultValue) must be of the shape { r: number; g: number, b: number, a: number; }.`,
       )
     }
+  }
 
-    // Clamp defaultValue components between 0 and 1
-    for (const component of ['r', 'g', 'b', 'a']) {
-      ;(defaultValue as $IntentionalAny)[component] = Math.min(
-        Math.max((defaultValue as $IntentionalAny)[component], 0),
-        1,
-      )
-    }
+  // Clamp defaultValue components between 0 and 1
+  const sanitized = {}
+  for (const component of ['r', 'g', 'b', 'a']) {
+    ;(sanitized as $IntentionalAny)[component] = Math.min(
+      Math.max((defaultValue as $IntentionalAny)[component], 0),
+      1,
+    )
   }
 
   return {
     type: 'rgba',
     valueType: null as $IntentionalAny,
-    default: decorateRgba(defaultValue),
+    default: decorateRgba(sanitized as Rgba),
     [propTypeSymbol]: 'TheatrePropType',
     label: opts?.label,
     sanitize: _sanitizeRgba,
@@ -306,14 +307,15 @@ const _sanitizeRgba = (val: unknown): Rgba | undefined => {
   }
 
   // Clamp defaultValue components between 0 and 1
+  const sanitized = {}
   for (const c of ['r', 'g', 'b', 'a']) {
-    ;(val as $IntentionalAny)[c] = Math.min(
+    ;(sanitized as $IntentionalAny)[c] = Math.min(
       Math.max((val as $IntentionalAny)[c], 0),
       1,
     )
   }
 
-  return valid ? decorateRgba(val as Rgba) : undefined
+  return valid ? decorateRgba(sanitized as Rgba) : undefined
 }
 
 const _interpolateRgba = (
--- a/theatre/core/src/propTypes/index.ts
+++ b/theatre/core/src/propTypes/index.ts
@@ -273,6 +273,35 @@ const _interpolateRgba = (
   progression: number,
 ): Rgba => {
   // TODO: Factor out all these utility functions
+  function linearSrgbToSrgb(rgba: Rgba) {
+    function compress(x: number) {
+      // This looks funky because sRGB uses a linear scale below 0.0031308 in
+      // order to avoid an infinite slope, while trying to approximate gamma 2.2
+      // as closely as possible, hence the branching and the 2.4 exponent.
+      if (x >= 0.0031308) return 1.055 * x ** (1.0 / 2.4) - 0.055
+      else return 12.92 * x
+    }
+    return {
+      r: compress(rgba.r),
+      g: compress(rgba.g),
+      b: compress(rgba.b),
+      a: rgba.a,
+    }
+  }
+
+  function srgbToLinearSrgb(rgba: Rgba) {
+    function expand(x: number) {
+      if (x >= 0.04045) return ((x + 0.055) / (1 + 0.055)) ** 2.4
+      else return x / 12.92
+    }
+    return {
+      r: expand(rgba.r),
+      g: expand(rgba.g),
+      b: expand(rgba.b),
+      a: rgba.a,
+    }
+  }
+
   function linearSrgbToOklab(rgba: Rgba) {
     let l =
       0.4122214708 * rgba.r + 0.5363325363 * rgba.g + 0.0514459929 * rgba.b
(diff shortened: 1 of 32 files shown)


theatre/core/src/propTypes/index.ts
```diff
@@ -234,6 +242,100 @@ const _interpolateNumber = (
   return left + progression * (right - left)
 }
 
+export const rgba = (
+  defaultValue: Rgba,
```
AriaMinaei, line 246: Maybe accept `undefined` here?

theatre/core/src/propTypes/index.ts
```diff
@@ -234,6 +242,100 @@ const _interpolateNumber = (
...
+
+    if (!valid) {
+      throw new Error(
+        `Argument defaultValue in t.rgba(defaultValue) must be of the shape { r: number; g: number, b: number, a: number; }.`,
+      )
+    }
+
+    // Clamp defaultValue components between 0 and 1
+    for (const component of ['r', 'g', 'b', 'a']) {
+      ;(defaultValue as $IntentionalAny)[component] = Math.min(
```
AriaMinaei, line 279: Caller should expect args to be immutable. So let's either throw a validation error, or create a new object.

theatre/core/src/propTypes/index.ts
```diff
@@ -234,6 +242,100 @@ const _interpolateNumber = (
   return left + progression * (right - left)
 }
 
+export const rgba = (
+  defaultValue: Rgba,
+  opts?: {
+    label?: string
+    sanitize?: Sanitizer<Rgba>
```
AriaMinaei, line 249: No need for a user-defined sanitizer I think

theatre/core/src/propTypes/index.ts
```diff
@@ -234,6 +242,100 @@ const _interpolateNumber = (
...
+    interpolate?: Interpolator<Rgba>
```
AriaMinaei, line 250: Also no need for a user-defined interpolator. They make sense on more general types like strings, whose value could have different formats and the user may want to interpolate between the components of those formats.

theatre/core/src/propTypes/index.ts
AriaMinaei: Ah I see it's just a remnant past code. We could just remove this

theatre/shared/src/utils/types.ts
```diff
@@ -11,7 +11,8 @@ export type SerializableMap<
   Primitives extends SerializablePrimitive = SerializablePrimitive,
 > = {[Key in string]?: SerializableValue<Primitives>}
 
-export type SerializablePrimitive = string | number | boolean
+// TODO: I'm sure we can make this type make more sense
+export type SerializablePrimitive = string | number | boolean | {}
```
AriaMinaei, line 15: I'm surprised `| {}` is not breaking anything. A `Primitive` is supposed to be, well, anything other than `Compound` (which itself includes `Enum`s). Adding `{}` to `Primitive` makes it accept anything iiuc. Maybe instead of `{}` we could hard-code `{r: number, g: number, b: number, a: number}` here until a better solution is found?

theatre/studio/src/StudioStore/StudioStore.ts
```diff
@@ -141,46 +142,62 @@ export default class StudioStore {
...
+                  return
+                }
+
+                const propAddress = {...root.address, pathToProp: path}
+
+                const trackId = get(sequenceTracksTree, path) as $FixMe as
+                  | SequenceTrackId
+                  | undefined
+                if (typeof trackId === 'string') {
+                  const propConfig = get(root.template.config.props, path) as
```
AriaMinaei, line 164: I think this wouldn't work for compound props. Because compound props' config looks like this `{type: 'compound', props: {}}`. `get()` wouldn't traverse the `props` prop. Use `getPropConfigByPath()` instead: https://github.com/AriaMinaei/theatre/blob/b97f51270e3cd3ab6c3fb607ef88939c0e89a0a5/theatre/shared/src/propTypes/utils.ts#L15

theatre/studio/src/StudioStore/StudioStore.ts
```diff
@@ -141,46 +142,62 @@ export default class StudioStore {
...
+                  if (propConfig?.sanitize) value = propConfig.sanitize(value)
+
+                  const seq = root.sheet.getSequence()
+                  seq.position = seq.closestGridPosition(seq.position)
+                  stateEditors.coreByProject.historic.sheetsById.sequence.setKeyframeAtPosition(
+                    {
+                      ...propAddress,
+                      trackId,
+                      position: seq.position,
+                      value: value as $FixMe,
```
AriaMinaei, line 176: Can't trust that user won't mutate `value`. Perhaps we should clone it?

theatre/studio/src/StudioStore/StudioStore.ts
```diff
@@ -141,46 +142,62 @@ export default class StudioStore {
...
+                      snappingFunction: seq.closestGridPosition,
+                    },
+                  )
+                } else {
+                  stateEditors.coreByProject.historic.sheetsById.staticOverrides.byObject.setValueOfPrimitiveProp(
+                    {...propAddress, value: value as $FixMe},
```
AriaMinaei, line 182: Also a clone here would be safer

theatre/studio/src/StudioStore/StudioStore.ts
AriaMinaei: ... because with the rgba type, `value` is no longer just `string | number | boolean`

theatre/studio/src/StudioStore/StudioStore.ts
```diff
@@ -141,46 +142,62 @@ export default class StudioStore {
...
+                const propAddress = {...root.address, pathToProp: path}
+
+                const trackId = get(sequenceTracksTree, path) as $FixMe as
+                  | SequenceTrackId
+                  | undefined
+                if (typeof trackId === 'string') {
+                  const propConfig = get(root.template.config.props, path) as
+                    | PropTypeConfig
+                    | undefined
+                  if (propConfig?.sanitize) value = propConfig.sanitize(value)
```
AriaMinaei, line 167: Now that we're decorating within sanitize, the value is not neccesarily serializable. This can remain for now, but maybe let's leave a `// TODO` that we need to `validate()` here instead of `sanitize()`.

theatre/studio/src/StudioStore/StudioStore.ts
```diff
@@ -202,33 +219,47 @@ export default class StudioStore {
...
+                if (typeof trackId === 'string') {
+                  stateEditors.coreByProject.historic.sheetsById.sequence.unsetKeyframeAtPosition(
+                    {
+                      ...propAddress,
+                      trackId,
+                      position: root.sheet.getSequence().positionSnappedToGrid,
+                    },
+                  )
+                } else {
+                  stateEditors.coreByProject.historic.sheetsById.staticOverrides.byObject.unsetValueOfPrimitiveProp(
```
AriaMinaei, line 246: We should account for when user runs `api.unset(object.props.color.r)`

theatre/studio/src/panels/DetailPanel/propEditors/RgbaPropEditor.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+  const stuff = useEditingToolsForPrimitiveProp<Rgba>(
+    pointerToProp,
+    obj,
+    propConfig,
+  )
+
+  const debouncedPermanentlySetValue = useMemo(
+    () =>
+      debounce((rgba: Rgba) => {
```
AriaMinaei, line 53: Probably an anti-pattern here. `temporarilySetValue` is there so we don't need to debounce and we can call it as often as we want. I think with debounce, there might be race conditions when sequence's position changes while the debouncer is waiting, which I think would create keyframe at the sequence's new position rather than at its original position

theatre/studio/package.json
```diff
@@ -28,6 +28,7 @@
   "dependencies": {
     "@theatre/dataverse": "workspace:*",
     "react": "^17.0.2",
+    "react-colorful": "^5.5.1",
```
AriaMinaei, line 31: Should we bundle this instead? we're bundling most other packages already.

AriaMinaei (changes requested):
Looks very good. Left a few comments

theatre/core/src/propTypes/index.ts
AndrewPrifer: Yeah, thought I'd cheat here for performance, good point

theatre/shared/src/utils/types.ts
AndrewPrifer: Yeah, ideally it would be a type that is any object where object.type !== 'compound', however that kind of type doesn't seem to be possible to construct with typescript.

The only trick that comes to mind is this, but it's ugly:
```ts
type SerializablePrimitive<T> = string | number | boolean | (T extends {type: 'compound'} ? never : T);

const badStuff = {
    type: 'compound',
    foo: 3,
} as const

const goodStuff = {
    type: 'literallyanythingelse',
    foo: 3,
} as const

function serializeStuff<T>(giveMeStuff: SerializablePrimitive<T>) {
    // ...
} 

serializeStuff(badStuff) // ⚠️
serializeStuff(goodStuff) // 💏
```

AndrewPrifer: I think for now I'll go with the hard-coded one, but I'll leave a comment to fix it.

theatre/studio/src/StudioStore/StudioStore.ts
AndrewPrifer: Cloning already happens both for the default value, and when interpolating, and I think this kind of stupid proofing should only happen on library boundaries anyway, when accepting, or when providing a value. Tbh the rgba object *is* cloned here when we call sanitize on it, so it does in fact happen here.

theatre/studio/src/StudioStore/StudioStore.ts
AndrewPrifer: You're right, let me see

AndrewPrifer: you were naughty too, this code is from you :D

theatre/studio/src/StudioStore/StudioStore.ts
AndrewPrifer: Oh, the API boundary *is* the set function, silly me. I'm going to clone right at the beginning.

theatre/studio/src/StudioStore/StudioStore.ts
AndrewPrifer: That decorating would have to do something really silly, like introduce circular references for that to happen 🤔 Or what kind of issue were you thinking exactly?

theatre/studio/package.json
AndrewPrifer: You mean bundle with core? This is a studio-only library so I don't see a reason.

theatre/studio/src/panels/DetailPanel/propEditors/RgbaPropEditor.tsx
AndrewPrifer: I'm not debouncing calls to `temporarilySetValue`, but to `permanentlySetValue`, similarly to how debouncedScrub does it, since the current color picker doesn't let us easily see when the picking started/ended. I agree that the ad-hoc nature of this is a little lazy, the less lazy alternative to this would be putting a debounced scrub method on `stuff`. Not saying there couldn't be bugs here, but those bugs wouldn't probably also be replicated in debouncedScrub.

theatre/studio/package.json
AriaMinaei: Bundle with studio I mean

theatre/studio/src/panels/DetailPanel/propEditors/RgbaPropEditor.tsx
AriaMinaei: Ah I see. Bad news is, `debouncedScrub()` also has that problem, and it's pretty annoying. We haven't shipped anything that uses `debouncedScrub()` (other than r3f, which isn't released yet) so I don't think users are encountering the problem.

I think it's a  good time to fix debouncedScrub and use it here, _or_ skip the debounce here. Otherwise => bad UX.

theatre/studio/src/StudioStore/StudioStore.ts
AriaMinaei: Let's say `sanitize()` (or decorate) returns a class instance. That's not straighforward to serialize.

theatre/studio/src/StudioStore/StudioStore.ts
```diff
@@ -30,6 +30,9 @@ import {isSheetObject} from '@theatre/shared/instanceTypes'
 import type {OnDiskState} from '@theatre/core/projects/store/storeTypes'
 import {generateDiskStateRevision} from './generateDiskStateRevision'
 import type {PropTypeConfig} from '@theatre/core/propTypes'
+import type {PathToProp} from '@theatre/shared/src/utils/addresses'
+import {getPropConfigByPath} from '../../../.temp/declarations/shared/src/propTypes/utils'
```
AriaMinaei, line 34: oops?

AriaMinaei: We need a lint rule to catch this.

theatre/studio/package.json
AndrewPrifer: Ah I understand. I think I'm gonna need help with configuring that. I see there's a bunch of stuff in theatre/package.json:devDependencies, is that something to do with this? The createBundle function defines externals, rather than packages to be bundled, so that should be fine I think, but my knowledge is lacking in this regard.

theatre/studio/src/panels/DetailPanel/propEditors/RgbaPropEditor.tsx
AndrewPrifer: Whatever bad ux we end up here is dwarfed by the bad UX of having an undo step at every color drag increment, so this seems like a fix, rather than a skip. Can you make me a minimal reproduction?

theatre/studio/package.json
AriaMinaei: Right. For now, we're just defining the dependencies on `theatre/package.json` unless we don't want those dependencies to get bundled, in which case we define them in `theatre/(core|studio)/package.json`.

- [ ] Let's add a quick note to document this

theatre/studio/src/StudioStore/StudioStore.ts
AndrewPrifer: Damn

theatre/studio/src/StudioStore/StudioStore.ts
AndrewPrifer: yep, well, we are going to overhaul the validate/sanitize/decode/decorate stuff anyway, so I'd leave it like this for now, will add a comment

theatre/studio/package.json
AndrewPrifer: Right, and what goes in dev deps and what in normal deps?

AndrewPrifer: Added notes about this.

theatre/studio/src/panels/DetailPanel/propEditors/RgbaPropEditor.tsx
AndrewPrifer: So, for now, fork react-colorful and add temporary/permanentlySetValue callbacks to it. We do need to fix debounced scrub soon though.

theatre/studio/src/StudioStore/StudioStore.ts
```diff
@@ -131,56 +134,77 @@ export default class StudioStore {
...
+                  stateEditors.coreByProject.historic.sheetsById.staticOverrides.byObject.setValueOfPrimitiveProp(
+                    {...propAddress, value: value as $FixMe},
+                  )
+                }
+              }
+
+              // If we are dealing with a compound prop, we recurse through its
+              // nested properties. The root prop with path.length === 0 is
+              // implicitly considered a compound prop.
+              if (path.length === 0 || propConfig?.type === 'compound') {
```
AriaMinaei, line 197: So, if propConfig is undefined, we shouldn't set that prop anyway, I think. Also, checking for `path.length === 0` looks brittle, because then `propConfig` _would_ be a compound.

AriaMinaei (changes requested):
So here is a code review. I'll also do a UX review a bit later. As for the code part, there is just a small change left to do.

theatre/studio/src/StudioStore/StudioStore.ts
AndrewPrifer: Oh, good catch, I was meaning remove the `?` because `propConfig` is never actually undefined, that was a remnant from some previous buggy code. `path.length === 0` is necessary because the root prop is an "implicit" compound prop, meaning it doesn't have `propConfig.type === 'compound'`, but should be treated as one.

PR #78, opened 2022-02-14
Fix r3f example not actually making the editable camera defined in the scene the defaul

AndrewPrifer:
The current example incorrectly uses <e.perspectiveCamera>, which doesn't make itself the default one, so the scene is actually still viewed through the default camera that r3f creates for us.

The correct solution is to create an editable perspective camera from drei's PerspectiveCamera, which exposes the option to make it the scene's default camera declaratively through the makeDefault prop.


danvoyce (approved):
LGTM 👍

AndrewPrifer:
@AriaMinaei it doesn't just work better, it works. :D The alternative would be passing the camera position to the Canvas element, but then it's not editable. If you think that's better, that's fine by me too. But this is the code we need to make cameras editable.

AndrewPrifer:
If you meant that we should make changes to the extension to fix this, we can only fix this in the way r3f fixed it: have our own version of drei. The problem is that three's camera doesn't let you make it the default camera declaratively, which Drei fixes by implementing the makeDefault prop.

AriaMinaei:
@AndrewPrifer Yeah I was thinking if there is a way to reduce boilerplate for the user. Not sure the implications of this, but can we ship drei's camera via dependency or hard-code it in?

AndrewPrifer:
> @AndrewPrifer Yeah I was thinking if there is a way to reduce boilerplate for the user. Not sure the implications of this, but can we ship drei's camera via dependency or hard-code it in?

I wouldn't hard-code it in, for the same reason r3f also doesn't ship drei's PerspectiveCamera. It is just too opinionated for a library that tries to align with vanilla threejs as much as possible. We could create theatre's version of an editable-drei library with objects like the editable Perspective camera being ready to just import, but I think that library needs a little more planning, and creating it would be a disproportionate effort considering we are talking about a single line of boilerplate.

PR #81, opened 2022-02-16
Feature/draggable area v2

danvoyce:
- [x] Adding the ability to drag infinitely when using the `DraggableArea` component
- [x] Implemented on `BasicNumberInput`
    - [x] Fixed issue that meant you had to blur the input before dragging 
- [x] Implemented on `LengthEditor`
    - [x] Fix issue that allowed `LengthIndicator` to be < 0 
- [x] General tidy / refactoring


theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/LengthIndicator/LengthIndicator.tsx
```diff
@@ -239,10 +240,12 @@ function useDragBulge(node: HTMLDivElement | null, props: IProps) {
           tempTransaction.discard()
           tempTransaction = undefined
         }
+
         tempTransaction = getStudio()!.tempTransaction(({stateEditors}) => {
+          const newLength = initialLength + delta
           stateEditors.coreByProject.historic.sheetsById.sequence.setLength({
             ...sheet.address,
-            length: initialLength + delta,
+            length: isGreaterThanZero(newLength) ? newLength : 0,
```
danvoyce, line 248: Fixing bug that allowed this to be dragged off the screen. Was there an issue for this?

theatre/studio/src/uiComponents/form/BasicNumberInput.tsx
```diff
@@ -275,14 +258,12 @@ const BasicNumberInput: React.FC<{
...
       onClick,
-      onFocus,
       onDragEnd,
       onDrag,
     }
-  }, [])
+  }, []) // TODO: fix the missing dependency warning
 
   let value =
     stateRef.current.mode !== 'editingViaKeyboard'
```
danvoyce, line 269: Does the user need the ability to edit more that 3 decimal places? 

Now that the user can drag whilst already focused (ie without clicking away first) it will display the extra decimal places for a split second (whilst `mode === 'editingViaKeyboard'). It's relatively minor, but a simple fix would be to only show a max of 3 decimals

https://user-images.githubusercontent.com/1253802/154534381-0900bcab-85d0-419a-b311-566aa8a1d995.mov

theatre/studio/src/uiComponents/form/BasicNumberInput.tsx
```diff
@@ -275,14 +270,23 @@ const BasicNumberInput: React.FC<{
...
       onInputKeyDown,
       onClick,
-      onFocus,
       onDragEnd,
       onDrag,
+      transitionToEditingViaKeyboardMode,
+      transitionToDraggingMode,
     }
-  }, [])
+  }, []) // TODO: fix the missing dependency warning
```
danvoyce, line 280: This will need quite a bit of refactoring so to reduce the noise in this PR I'll do this separately

theatre/studio/src/uiComponents/form/BasicNumberInput.tsx
```diff
@@ -133,17 +131,22 @@ const BasicNumberInput: React.FC<{
...
     const inputChange = (e: React.ChangeEvent) => {
       const target = e.target as HTMLInputElement
       const {value} = target
       const curState = stateRef.current as IState_EditingViaKeyboard
 
-      stateRef.current = {...curState, currentEditedValueInString: value}
+      stateRef.current = {
+        ...curState,
+        currentEditedValueInString: value,
+      }
```
danvoyce, line 147: @AriaMinaei should the value when `editingViaKeyboard` take into consideration the `range` prop?

theatre/studio/src/uiComponents/form/BasicNumberInput.tsx
danvoyce: In order to fix a couple of weird bugs I ended up refactoring as part of this PR, so not follow up PR now.

AndrewPrifer:
UX issues (I know, some of these might have been inherited):
- Scrubbing is also enabled when the input is focused, which prevents making selections even when focused. If we really want to keep the whole-area dragging, I think disallowing scrubbing in focus is the lesser evil, since this essentially disables a standard feature and replaces it with an unexpected edit.
  - If we really want this, we need to display the drag indicator rather than the default editing cursor when in focus too.
- Scrubbing an input focuses it. I think focus should be preserved.
- When acquiring pointer lock on safari, the "press ESC to etc." notification pushes down the entire page which even makes me motion sick when it happens a lot, and scrubbing happens a lot. I'd disable this on safari.
- Scrubbing only makes sense in conjunction with a certain precision. Sometimes we might not care about decimals, sometimes all we care about is decimals, sometimes we just want to scrub the tens. This precision should be configurable for each input. Additionally, we should provide modifier keys for adjusting the precision up/down.
- I'd round the input to the precision of the scrubbing, under the assumption that when scrubbing, the user doesn't care about preserving higher-precision digits.

PR #89, opened 2022-03-01
API proposal: add a way to get a project state pointer

felthy:
We've encountered two use cases so far where we've wanted to react to project state changes:

1. When building a UI that uses keyframe data, need to be able to react to keyframe changes including when easing curves are changed by the author
2. To auto-save while authors are using the studio, need to react to all changes anywhere in the state (except metadata like `definitionVersion` and `revisionHistory`)

Knowing that `val()` and `onChange()` are already exposed from core, all we need is a pointer to the top of the “historic” state. I considered adding a `state` property to `IProject`, similar to how `ISheetObject` exposes `props`; however the fact that this feature will only be useful in the context of the studio led me to follow the example of `createContentOfSaveFile()` by adding `getProjectState()` to `IStudio`.

### DIFF
--- a/theatre/studio/src/TheatreStudio.ts
+++ b/theatre/studio/src/TheatreStudio.ts
@@ -1,7 +1,7 @@
 import type {IProject, ISheet, ISheetObject} from '@theatre/core'
 import studioTicker from '@theatre/studio/studioTicker'
 import type {IDerivation, Pointer} from '@theatre/dataverse'
-import {prism} from '@theatre/dataverse'
+import {prism, val} from '@theatre/dataverse'
 import SimpleCache from '@theatre/shared/utils/SimpleCache'
 import type {$IntentionalAny, VoidFn} from '@theatre/shared/utils/types'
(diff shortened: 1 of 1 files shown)


vezwork:
Hi @felthy thanks for the PR! I really appreciate the inline documentation with usage examples and your explanation of your use cases.

Code-wise this PR looks good to me. I will have to ask @AriaMinaei if this addition makes sense in the context of our plans for the API. @AriaMinaei what do you think? Can we accept this change, or do we want another way to make this possible?

vezwork:
quick update: we are discussing this internally and we will keep you updated as we make progress. I can't promise any sort of timeline but I wanted to keep you updated on the status of the PR from our perspective.

felthy:
Hi @AriaMinaei, yes we are actively using this in our fork for the two use cases in the PR description. Any API that can work for those use cases will be sufficient - this PR is the way I came up with that meets our needs, and also felt in line with the way theatre is already doing things (e.g. `val()` was already exposed publicly).

PR #92, opened 2022-03-02
Make it possible to move the playhead to an exact position

fulopkovacs:
This feature should work like this:
- the user clicks on the playhead in the Time Sequence Editor
- a popup appears with a number input, where the user can type in the exact frame they want to jump to

### DIFF
--- a/theatre/studio/src/panels/SequenceEditorPanel/RightOverlay/PlayheadPositionPopover.tsx
+++ b/theatre/studio/src/panels/SequenceEditorPanel/RightOverlay/PlayheadPositionPopover.tsx
@@ -1,0 +1,89 @@
+import styled from 'styled-components'
+import type {SequenceEditorPanelLayout} from '@theatre/studio/panels/SequenceEditorPanel/layout/layout'
+import {usePrism} from '@theatre/react'
+import type {BasicNumberInputNudgeFn} from '@theatre/studio/uiComponents/form/BasicNumberInput'
+import BasicNumberInput from '@theatre/studio/uiComponents/form/BasicNumberInput'
+import {propNameText} from '@theatre/studio/panels/DetailPanel/propEditors/utils/SingleRowPropEditor'
+import {useLayoutEffect, useMemo, useRef} from 'react'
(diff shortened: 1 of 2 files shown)


theatre/studio/src/panels/SequenceEditorPanel/RightOverlay/PlayheadPositionPopover.tsx
```diff
@@ -0,0 +1,89 @@
...
+  onRequestClose: () => void
+}> = ({layoutP, onRequestClose}) => {
+  const sheet = val(layoutP.sheet)
+  const sequence = sheet.getSequence()
+
+  const fns = useMemo(() => {
+    let tempPosition: number | undefined
+
+    return {
+      temporarilySetValue(newPosition: number): void {
```
AriaMinaei, line 44: This is very good. My only suggestion is to make the number input react to each keystroke so the user can see the playhead move as they type the number in. Honestly I don't know how that'd be helpful to the user, but intuitively feel it'd be useful, and since it's easy to implement and maintain, I'd suggest to implement it.

It should be straightforward, with the `temporarilySetValue()` and `discardTemporaryValue()` functions.

AriaMinaei (changes requested):
Great work! I just added a single suggestion.

theatre/studio/src/panels/SequenceEditorPanel/RightOverlay/PlayheadPositionPopover.tsx
fulopkovacs: That's a good suggestion, I implemented it. 

By the way, when I was testing the feature I  had an idea: there should be a way to center the playhead in the Time Sequence Editor. Sometimes when I zoom in too much I lose the playhead, so these times it'd be nice to just click a button and put the playhead in the center (the playhead's position _in time_ would not change, only the position in the editor's panel). When we jump to an exact point in time we would also want the follow the playhead in my opinion. 

That said, I would put this feature in another PR, since this feature is already useful in its current state. :relaxed:

PR #94, opened 2022-03-04
Add focus range

fulopkovacs:


fulopkovacs:
**Status:** WIP, not ready to merge!

I have reached the point where we have a messy Focus Range with dragable and movable handles for the start and end points of the range. It also controls the playback once you moved the handles AND restarted the playback. I have some thoughts on how to proceed, but a review now would save us a few other ones in the future. :laughing: A Slack-huddle/Zoom call would be appreciated, since I have a few questions.

AriaMinaei:
Looks like the drag gesture has a drift in FF depending on zoom level, but not in Chrome.

https://user-images.githubusercontent.com/593118/157208109-b71ba93c-adac-4962-8760-d0eac4f3b831.mp4

fulopkovacs:
Current state of the PR:
- [x] fix the height of the highlighted part of the DopeSheet
- [x] minimal size
- [x] double click for max size
- [ ] replace the current color of `focusRangeStrip` when dragging
- [ ] replace  the current color of  `focusRangeThumb` when dragging
- [ ] clean up the focus Range files
- [ ] tooltips when dragging
- [ ] refactor the  `focusRangeTheme` object
- [ ] [bugfix] Don't highlight the range on hover when it's disabled

AriaMinaei (changes requested):
This is great! Just a couple suggestions:

1. Review the code comments and comment anything that's not obvious (like `sequence.playInDynamicRange()`
2. Make the thumbs snappable.

fulopkovacs:
Current status after the 1st review:
- [x] Make the thumbs snapable
- [ ] Add comments to the parts that are harder to read

PR #99, opened 2022-03-15
Change ESM builds for R3F extension so the components are treeshakeable 🌲

### LINKED ISSUE #97: theatre/r3f: Bundle includes theatre/studio
When building a project with Webpack or Vite using `@theatre/r3f`, it looks like `@theatre/studio` is being pulled into the build unexpectedly. I've included a simple 'hello world' project to reproduce the issue in the .zip below, and my `index.js` is:

```js
import { getProject } from '@theatre/core';
import { editable as e } from '@theatre/r3f';

console.log('index.js::init', { getProject, e });

if ("__TheatreJS_StudioBundle" in window) {
	console.log('❌ Unexpected @theatre/studio bundle found.');
} else {
	console.log('✅ @theatre/studio not included.');
}
```

My project is using Vite, but the test case is a bit simpler using Webpack.

[Shared.zip](https://github.com/theatre-js/theatre/files/8218352/Shared.zip)

```
yarn
yarn build
npx serve dist
```

juniorxsound:
## What?
Right now when `esbuild`ing the R3F extension everything is bundled into a single file for both ESM and CJS builds (including `import` statements to `@theatre/studio` which are only required by some of the components). This PR moves to building an ESM version without bundling all components to a single file, that way separate components can be imported by the user without fear that `studio` would be included in your bundle when not needed (e.g when importing just `editable` without `studio` in the project).

### Before
Here is a snapshot of the current build output (`esm` target builds a single `index.mjs` file with everything inlined)
Screen Shot 2022-03-15 at 4 27 41 PM

### After
Screen Shot 2022-03-15 at 4 29 25 PM
Each component is built into JS but kept separated by turning  `esbuild`'s `bundle` prop off - that way components can be imported individually without having to import the whole `index.mjs` dependencies.

Closes #97

### DIFF
--- a/packages/r3f/devEnv/build.ts
+++ b/packages/r3f/devEnv/build.ts
@@ -2,6 +2,7 @@ import * as path from 'path'
 import {build} from 'esbuild'
 import type {Plugin} from 'esbuild'
 import {existsSync, mkdirSync, writeFileSync} from 'fs'
+import {globPlugin} from 'esbuild-plugin-glob'
 
 const externalPlugin = (patterns: RegExp[]): Plugin => {
   return {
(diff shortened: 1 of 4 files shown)


nbartzokas (approved):
LGTM!

AriaMinaei:
Thanks for the PR @juniorxsound and @nbartzokas!

@AndrewPrifer is fixing up the build system via #101 so I think this PR would be reviewed as part of that task.

AndrewPrifer:
This is now (finally, yay) addressed by #129, along with various other issues. Thanks for your contribution @juniorxsound, we ended up using your solution and even removing the cjs build altogether.

PR #107, opened 2022-03-24
Fix keyframe pasting to respect interpolation graph

vezwork:
## Demo
Before the fix:
1. the curve is slightly horizontally offset from the keyframes 
2. the green y curve is copied from x, but does not match x's curve
3. the curve does not properly display default values when the keyframe has a type that does not match the track (not pictured)
Screen Shot 2022-03-24 at 4 46 30 PM
after the fix:

1. the curve is correctly aligned with the keyframes
2. the green y is copied from x, and *does* match x
3. the curve correctly shows default values (not pictured)
Screen Shot 2022-03-24 at 4 50 39 PM

## Note
- also adds 'copy keyframe' to a single keyframe's context menu. Missed this on the first pass.

### DIFF
--- a/theatre/shared/src/propTypes/utils.ts
+++ b/theatre/shared/src/propTypes/utils.ts
@@ -30,8 +30,8 @@ export function getPropConfigByPath(
 }
 
 /**
- * @param value An arbitrary value. May be matching the prop's type or not
- * @param propConfig The configuration object for a prop
+ * @param value - An arbitrary value. May be matching the prop's type or not
+ * @param propConfig - The configuration object for a prop
  * @returns value if it matches the prop's type, otherwise returns the default value for the prop
  */
 export function valueInProp(value: unknown, propConfig: PropTypeConfig) {
--- a/theatre/shared/src/propTypes/utils.ts
+++ b/theatre/shared/src/propTypes/utils.ts
@@ -28,3 +28,12 @@ export function getPropConfigByPath(
 
   return getPropConfigByPath(sub, rest)
 }
+
+export function valueInProp(value: unknown, propConfig: PropTypeConfig) {
+  const sanitizedVal = propConfig.sanitize!(value)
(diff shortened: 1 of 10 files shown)


AriaMinaei (approved):

AriaMinaei:
The `!` makes this function's type signature unreliable. 

image

theatre/shared/src/propTypes/utils.ts
```diff
@@ -28,3 +36,23 @@ export function getPropConfigByPath(
...
+ * @param value - An arbitrary value. May be matching the prop's type or not
+ * @param propConfig - The configuration object for a prop
+ * @returns value if it matches the prop's type, otherwise returns the default value for the prop
+ */
+export function valueInProp<PropValueType>(
+  value: unknown,
+  propConfig: IBasePropType<PropValueType>,
+): PropValueType {
+  const sanitize = propConfig.sanitize
+  if (!sanitize) return propConfig.default
```
AriaMinaei, line 50: Hmm, we probably need to make `sanitize` mandatory. Otherwise, the prop's value would always be its default value. But that's for another PR.

AriaMinaei (changes requested):

theatre/shared/src/propTypes/utils.ts
AriaMinaei: So I'd say that, for now, let's return the given value rather than the default value (which does match the assumptions made in `core/src/propTypes/index.ts`.

In another PR (which I'm working on), `sanitize()` will be mandatory, so this would not be a problem anymore.

vezwork: Thanks, this confirms and fills in my understanding. I am glad `sanitize` will be mandatory. I have modified `valueInProp` with this info in mind. I have also reverted some of my other extra typing changes so as not to encode extra assumptions before your PR is merged.

AriaMinaei (approved):

PR #114, opened 2022-03-30
theatre/r3f: Allow physical lighting in snapshot viewport

donmccurdy:
Related:

- https://github.com/pmndrs/react-three-fiber/issues/2127

I'd encourage any new project to enable `renderer.physicallyCorrectLights=true`, but it's difficult to change the default in three.js or R3F at this point. There is no downside to physical lighting (you can still reduce decay for non-physical light range if needed), we're just stuck with it off for backward-compatibility.  In order to at least support the more modern and correct workflow, could we expose this as a viewport option?

*This contribution is funded by The New York Times.*


vezwork:
@AndrewPrifer should this go in?

AndrewPrifer:
Hey @vezwork thanks for the reminder, completely forgot about this. There are some conflicts to resolve, gonna merge it later today.

PR #119, opened 2022-04-06
Update GET_STARTED.md

colelawrence:


packages/dataverse/docs/GET_STARTED.md
```diff
@@ -262,12 +262,12 @@ This would require the derivation to do the following steps:
...
 
 ```typescript
-import {prism} from '@theatre/dataverse'
+import { prism } from '@theatre/dataverse'
 
 const mousePositionD = prism(() => {
   // Create an internal state (`pos`) where the position of the mouse
   // will be stored, and a function that updates it (`setPos`)
-  const [pos, setPos] = prism.state<[x: number, y: number]>('pos', [0, 0])
+  const [pos, setPos] = prism.state('pos', { x: 0, y: 0 })
```
colelawrence, line 270: By using a structure, it ensures that TypeScript will be able to fully infer all types.

colelawrence:
Recommend squash / merge

Issue #28, opened 2021-10-01
Ability to select a predefined easing between keyframes (e.g. 'linear')


jmyrland:
Feature request based on [this thread](https://discord.com/channels/870988717190426644/889485298575212604/893590264806006804).

The default easing between two keyframes is by default set to `ease-in-out` after inserting a new keyframe. This is fine for most cases, but in some cases it might be useful to provide a way to select another easing between to keyframes, like `linear`.

E.g. when right-clicking the easing-curve between 2 keyframes, it would be nice to present a menu of predefined easing-curves to provide a quick way to set a curve without having to manually adjust the handles.

The prime example for this, is trying to get a perfectly linear curve between two keyframes. Today, if the keyframes are spread far apart, it is quite tricky to get the curve perfectly linear. One workaround is to adjust the state.json manually.

It would be nice to support a predefined set of easing-curves, like how [css have a set of predefined easing functions](https://developer.mozilla.org/en-US/docs/Web/CSS/easing-function):
- `linear [cubic-bezier(0.0, 0.0, 1.0, 1.0)]`
- `ease [cubic-bezier(0.25, 0.1, 0.25, 1.0)]`
- `ease-in [cubic-bezier(0.42, 0.0, 1.0, 1.0)]`
- `ease-in-out [cubic-bezier(0.42, 0.0, 0.58, 1.0)]`
- `ease-out [cubic-bezier(0.0, 0.0, 0.58, 1.0)]`

michaeltheory:
Hoping to be a built helpful and provide a list of cubic beziers we use for our code-based tweening

`{name: 'easeOutCubic', curve: 'cubic-bezier(0.215, 0.610, 0.355, 1.000)'},
{name: 'easeOutQuad', curve: 'cubic-bezier(0.250, 0.460, 0.450, 0.940)'},
{name: 'easeOutQuart', curve: 'cubic-bezier(0.165, 0.840, 0.440, 1.000)'},
{name: 'easeOutQuint', curve: 'cubic-bezier(0.230, 1.000, 0.320, 1.000)'},
{name: 'easeOutSine', curve: 'cubic-bezier(0.390, 0.575, 0.565, 1.000)'},
{name: 'easeOutExpo', curve: 'cubic-bezier(0.190, 1.000, 0.220, 1.000)'},
{name: 'easeOutCirc', curve: 'cubic-bezier(0.075, 0.820, 0.165, 1.000)'},
[...]

PR #122, opened 2022-04-13
feature/single curve editor

vezwork:
## Changes
- Add curve editor popover
- Add `usePointerCapturing` provider

## Demo

NEW
https://user-images.githubusercontent.com/11082236/164542544-f1f66de2-f62e-44dd-b4cb-05b5f6e73a52.mp4

OLD
Screen Shot 2022-04-14 at 4 17 17 PM


theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/CurveEditorPopover.tsx
```diff
@@ -233,225 +183,128 @@ const CurveEditorPopover: React.FC<
...
+            onKeyDown={onEasingOptionKeydown(preset)}
+            ref={optionsRef.current[preset.label]}
+            onClick={() => {
+              fns.permanentlySetValue(preset.value)
+              //props.onRequestClose()
+            }}
+            // Temporarily apply on hover
+            onMouseOver={() => {
+              // When previewing with hover, we don't want to set the filter too
+              //fns.temporarilySetValue(preset.value)
```
colelawrence, line 295: Change: This comment was previously attached to an uncommented following line. But now the following line is commented so it has a changed meaning.

If you're not sure about removing this dead code, and want to keep it, then at least make sure the comment look like:
```ts
// // When previewing with hover, we don't want to set the filter too
// fns.temporarilySetValue(preset.value)
```

So it is obvious that the entire code with the comment is commented out.

Alternatively, comment out the entire `onMouseOver` value with a comment describing why it might be interesting to keep the code around.

Perhaps just delete the code here entirely and capture any uncertainty about this in a different nearby comment capturing the thought process / development journey.

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/CurveEditorPopover.tsx
```diff
@@ -233,225 +183,128 @@ const CurveEditorPopover: React.FC<
...
+            }}
+            onMouseOut={() => {
+              //fns.discardTemporaryValue()
+            }}
```
colelawrence, line 299: Also perhaps delete / document per above suggestion

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/CurveEditorPopover.tsx
```diff
@@ -233,225 +183,128 @@ const CurveEditorPopover: React.FC<
...
+          />
+        ))}
+      </OptionsContainer>
+      <CurveEditorContainer>
+        <CurveSegmentEditor {...props}></CurveSegmentEditor>
```
colelawrence, line 304: Nit: If you have no children, you can just `<CurveSegmentEditor {...props}/>`

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/EasingOption.tsx
```diff
@@ -0,0 +1,68 @@
...
+                        dangerouslySetInnerHTML={{
+                          __html: fuzzySort.highlight(
+                            presetSearchResults[index] as any,
+                          )!,
+                        }}
+                      />
+                    ) : (
+                      preset.label
+                    )}
+                  </span> */}
```
colelawrence, line 63: Change: Add comment for context (in the spirit of above review comments) or delete entirely.

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/SVGCurveSegment.tsx
```diff
@@ -0,0 +1,84 @@
...
+const SVGCurveSegment: React.FC<IProps> = (props) => {
+  const {easing} = props
+  if (easing) {
+    return (
+      <svg
+        height="100%"
+        width="100%"
+        viewBox={`${-SVG_PADDING} ${-SVG_PADDING} ${1 + SVG_PADDING * 2} ${
+          1 + SVG_PADDING * 2
+        }`}
```
colelawrence, line 23: Change: Add docs to describe what value you're trying to construct here and why.

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/SVGCurveSegment.tsx
```diff
@@ -0,0 +1,84 @@
...
+        fill="none"
+        xmlns="http://www.w3.org/2000/svg"
+      >
+        <line
+          x1="0"
+          y1="1"
+          x2={easing[0]}
+          y2={1 - easing[1]}
```
colelawrence, line 31: Change: Please name these variables to create self describing code at the top of this file–separate from the JSX declaration.

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/SVGCurveSegment.tsx
```diff
@@ -0,0 +1,84 @@
...
+          r={0.2}
+          fill={CONTROL_HITZONE_COLOR}
+        />
+
+        <path
+          d={`M0 1 C${easing[0]} ${1 - easing[1]} ${easing[2]} 
+      ${1 - easing[3]} 1 0`}
+          stroke={CURVE_COLOR}
+          strokeWidth="0.08"
+        />
```
colelawrence, line 62: Same as above recommendations: name these variables.

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/SVGCurveSegment.tsx
```diff
@@ -0,0 +1,84 @@
...
+const CONTROL_COLOR = '#4f4f4f'
+const CONTROL_HITZONE_COLOR = 'rgba(255, 255, 255, 0.1)'
+
+type IProps = {
+  easing: [number, number, number, number] | null
+}
+
+const SVGCurveSegment: React.FC<IProps> = (props) => {
+  const {easing} = props
+  if (easing) {
```
colelawrence, line 16: Change: Flip this pattern with [the "return early" style](https://medium.com/swlh/return-early-pattern-3d18a41bba8).

i.e.

```ts
if (!easing) {
  return <></>
}

// declare variables

return (
  <svg
    ...
```

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/SVGCurveSegment.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+const SVG_PADDING = 1
+const SVG_CIRCLE_RADIUS = 0.1
+
+const CURVE_COLOR = '#858585'
+const CONTROL_COLOR = '#4f4f4f'
+const CONTROL_HITZONE_COLOR = 'rgba(255, 255, 255, 0.1)'
+
+type IProps = {
+  easing: [number, number, number, number] | null
```
colelawrence, line 11: Change: Use shared type definition for this easing value as mentioned before.

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/shared.ts
```diff
@@ -0,0 +1,99 @@
+import {useMemo, useRef, useState} from 'react'
+
+// n mod m
+export function mod(n: number, m: number) {
+  return ((n % m) + m) % m
+}
```
colelawrence, line 6: Suggestion: This is only used in one file, I would strongly recommend keeping it in that one file, so it isn't accidentally imported and reused by another part of the codebase that could misunderstand it's purpose.

Suggestion: If applicable, document or paste link to reference of what this is useful for in regards to the context you want to use it for.

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/shared.ts
```diff
@@ -0,0 +1,99 @@
...
+  {label: 'Quint In', value: '0.755, 0.050, 0.855, 0.060'},
+  {label: 'Quint Out', value: '0.230, 1, 0.320, 1'},
+
+  {label: 'Sine In Out', value: '0.445, 0.050, 0.550, 0.950'},
+  {label: 'Sine In', value: '0.470, 0, 0.745, 0.715'},
+  {label: 'Sine Out', value: '0.390, 0.575, 0.565, 1'},
+
+  {label: 'Ease Out In', value: '.42, 0, .58, 1'},
+  {label: 'Linear', value: '0.5, 0.5, 0.5, 0.5'},
+]
```
colelawrence, line 66: Question: Why are these presets fixed precision 3, while the to string code is `toFixed(2)`?

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/shared.ts
```diff
@@ -0,0 +1,99 @@
...
+function useRefreshableMemo<T>(
+  fn: (refresh: () => void) => T,
+  additionalDeps: any[],
+): T {
+  const [version, setVersion] = useState(0)
+  return useMemo(
+    () => fn(() => setVersion(version + 1)),
+    [version, ...additionalDeps],
+  )
+}
```
colelawrence, line 99: Change: This is not used anywhere, so delete.

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/shared.ts
```diff
@@ -0,0 +1,99 @@
...
+  fn: (setFreeze: (isFrozen: boolean) => void) => T,
+  deps: any[],
+): T {
+  const [isFrozen, setFreeze] = useState<boolean>(false)
+  const freezableDeps = useRef(deps)
+
+  if (!isFrozen) freezableDeps.current = deps
+
+  return useMemo(() => fn(setFreeze), freezableDeps.current)
+}
```
colelawrence, line 86: Idea: Might be good to instead of having a `shared.ts` file, to have a `shared/useFreezableMemo.ts` file since these items are not actually related to each other.

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/Dot.tsx
```diff
@@ -18,9 +18,9 @@ import {useCssCursorLock} from '@theatre/studio/uiComponents/PointerEventsHandle
 import SnapCursor from './SnapCursor.svg'
 import selectedKeyframeIdsIfInSingleTrack from '@theatre/studio/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/selectedKeyframeIdsIfInSingleTrack'
 
-export const dotSize = 6
-const hitZoneSize = 12
-const snapCursorSize = 34
+export const DOT_SIZE = 6
+const HIT_ZONE_SIZE = 12
+const SNAP_CURSOR_SIZE = 34
```
colelawrence, line 23: Change: Add initial doc comments for what these constants represent and what unit they are in (I assume dp / px?)

Doesn't need to be the best comments ever, just something, since you're probably more aware of what they are than I can be by simply reading the identifier names.

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/Dot.tsx
```diff
@@ -44,12 +44,12 @@ const Square = styled.div<{isSelected: boolean}>`
 
   z-index: 1;
   pointer-events: none;
-  ${(props) => dims(props.isSelected ? dotSize : dotSize)}
+  ${(props) => dims(props.isSelected ? DOT_SIZE : DOT_SIZE)}
```
colelawrence, line 47: Change: Give more descriptive name to `dims` and if not already, perhaps document it.

Question: If `dims` is a pure function, then can we just `${dims(DOT_SIZE)}`?

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/Dot.tsx
```diff
@@ -59,10 +59,10 @@ const HitZone = styled.div`
...
     &:hover:after {
       position: absolute;
-      top: calc(50% - ${snapCursorSize / 2}px);
-      left: calc(50% - ${snapCursorSize / 2}px);
-      width: ${snapCursorSize}px;
-      height: ${snapCursorSize}px;
+      top: calc(50% - ${SNAP_CURSOR_SIZE / 2}px);
+      left: calc(50% - ${SNAP_CURSOR_SIZE / 2}px);
+      width: ${SNAP_CURSOR_SIZE}px;
+      height: ${SNAP_CURSOR_SIZE}px;
```
colelawrence, line 65: Nit: Instead of doing a `/ 2` every time, perhaps have a `SNAP_CURSOR_SIZE_HALF` value.

Random and totally just trivia: in microbenchmarks, floating point division is a magnitude more expensive for the CPU than multiplication.

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/Dot.tsx
```diff
@@ -75,7 +75,7 @@ const HitZone = styled.div`
   }
 
   &:hover + ${Square}, &.beingDragged + ${Square} {
-    ${dims(dotSize + 5)}
+    ${dims(DOT_SIZE + 5)}
```
colelawrence, line 78: Question: What is `+ 5`?
This kinda looks like something else that should be a constant with a name.

theatre/studio/src/uiComponents/simpleContextMenu/RightClickMenu/RightClickMenu.tsx
```diff
@@ -45,6 +45,7 @@ const RightClickMenu: React.FC<{
   items: IContextMenuItem[] | (() => IContextMenuItem[])
   rightClickPoint: {clientX: number; clientY: number}
   onRequestClose: () => void
+  keepOpen?: boolean
```
colelawrence, line 48: Change: Add inline doc comment for why `keepOpen` is useful, and what it prevents.

theatre/studio/src/uiComponents/simpleContextMenu/RightClickMenu/RightClickMenu.tsx
```diff
@@ -80,7 +81,7 @@ const RightClickMenu: React.FC<{
 
     const onMouseMove = (e: MouseEvent) => {
       if (
-        e.clientX < pos.left - pointerDistanceThreshold ||
+        (!props.keepOpen && e.clientX < pos.left - pointerDistanceThreshold) ||
```
colelawrence, line 84: Suggest: I strongly recommend trying to use another early return style here for each case. And generally cleaning this up, because it looks very strange that `props.keepOpen` is only affecting one part of the `||` expression chain.

theatre/studio/src/uiComponents/simpleContextMenu/RightClickMenu/RightClickMenu.tsx
colelawrence: I might of missed it, but please double check that this `keepOpen` is actually being used for something.

theatre/studio/src/uiComponents/useDrag.ts
```diff
@@ -19,9 +26,14 @@ export type UseDragOpts = {
   /**
    * Called at the start of the gesture. Mind you, that this would be called, even
    * if the user is just clicking (and not dragging). However, if the gesture turns
-   * out to be a click, then onDragEnd(false) will be called. Otherwise,
+   * out to be a click, then `onDragEnd(false)` will be called. Otherwise,
    * a series of `onDrag(dx, dy, event)` events will be called, and the
    * gesture will end with `onDragEnd(true)`.
+   *
+   *
+   * @return
```
colelawrence, line 34: Perhaps the keyword is actually `@returns`. Check other places in codebase...

colelawrence (changes requested):
See comments.

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/CurveEditorPopover.tsx
```diff
@@ -233,225 +183,128 @@ const CurveEditorPopover: React.FC<
...
+  const onEasingOptionKeydown =
+    (preset: {label: string; value: string}) =>
+    (e: KeyboardEvent<HTMLInputElement>) => {
+      if (e.key === 'ArrowRight') moveHighlightHorizontal(1)
+      if (e.key === 'ArrowLeft') moveHighlightHorizontal(-1)
+      if (e.key === 'ArrowUp') moveHighlightVertical(-1)
+      if (e.key === 'ArrowDown') moveHighlightVertical(1)
+      if (e.key === 'Escape') props.onRequestClose()
+      if (e.key === 'Enter') fns.permanentlySetValue(preset.value)
+    }
```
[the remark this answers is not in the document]
vezwork: I am not aware of any such considerations but it definitely makes sense to add keybindings at some point.

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/Dot.tsx
vezwork: I just renamed these using constant styling but I have not touched this code. I looked into a bit and they are in px units. Added _PX postfix.

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/Dot.tsx
vezwork: I will make this change because yes this makes sense

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/Dot.tsx
vezwork: Interesting trivia. I will elect not to modify this because the division is only being performed once when loading.

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/Dot.tsx
vezwork: Added a constant based on a brief look at the code

theatre/studio/src/uiComponents/simpleContextMenu/RightClickMenu/RightClickMenu.tsx
vezwork: it is dead. Removed. ty.

theatre/studio/src/uiComponents/simpleContextMenu/RightClickMenu/RightClickMenu.tsx
vezwork: this code was dead and the change is removed.

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/SVGCurveSegment.tsx
```diff
@@ -0,0 +1,96 @@
...
+      <circle
+        cx={RIGHT_CONTROL_POINT[0]}
+        cy={RIGHT_CONTROL_POINT[1]}
+        r={0.2}
+        fill={CONTROL_HITZONE_COLOR}
+      />
+
+      <path
+        d={`M0 1 C${LEFT_CONTROL_POINT[0]} ${LEFT_CONTROL_POINT[1]} ${RIGHT_CONTROL_POINT[0]} 
+      ${RIGHT_CONTROL_POINT[1]} 1 0`}
```
colelawrence, line 72: Suggestion: Consider naming and moving this closer to the other vars above as well
```ts
const curvePathDAttrValue = `M0 1 C${LEFT_CONTROL_POINT[0]} ${LEFT_CONTROL_POINT[1]} ${RIGHT_CONTROL_POINT[0]} 
      ${RIGHT_CONTROL_POINT[1]} 1 0`
```

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/colors.ts
```diff
@@ -0,0 +1,5 @@
+export const COLOR_POPOVER_BACK = 'black'
+
+export const COLOR_BASE = '#292929'
+
+export const COLOR_FOCUS_OUTLINE = '#f78104'
```
colelawrence, line 5: Thinking out loud: Separating colors out like this feels a bit strange to me. I'm going to be keeping my eye on this kind of pattern and whether there's something interesting we'd want to move towards with custom theming / custom branding for users.

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/shared.ts
```diff
@@ -0,0 +1,82 @@
+/*
+ * Cubic bezier curves can be defined by two 2d points in a 4-tuple.
+ * The 4-tuple defines the start control point x1,y1 and the end control point x2,y2.
+ * x1 should always be >= 0 and x2 must be <= 1.
+ * It is assumed that the start of the curve is fixed at 0,0 and the end is fixed at 1,1.
+ * to get a feel for it: https://cubic-bezier.com/
+ */
+export type CubicBezierPoints = [x1: number, y1: number, x2: number, y2: number]
```
colelawrence: Praise: Thanks for breaking this out and adding the illustrious comment!

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/shared.ts
```diff
@@ -0,0 +1,89 @@
...
+export type CubicBezierHandles = [
+  x1: number,
+  y1: number,
+  x2: number,
+  y2: number,
+]
+
+export function cssCubicBezierArgsFromHandles(points: CubicBezierHandles) {
+  return points.map((p) => p.toFixed(2)).join(', ')
+}
```
colelawrence, line 17: Change: It was pointed out this should be `p.toFixed(3)`, IIUC.

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/shared.ts
```diff
@@ -0,0 +1,89 @@
...
+  easing1: CubicBezierHandles | null | undefined,
+  easing2: CubicBezierHandles | null | undefined,
+) {
+  if (!easing1 || !easing2) return false
+  let totalDiff = 0
+  for (let i = 0; i < 4; i++) {
+    totalDiff += Math.abs(easing1[i] - easing2[i])
+  }
+  return totalDiff < SIMILARITY_CUTOFF
+}
```
colelawrence, line 89: Change: Add doc comment about how this is being used. Potentially require that the similarity cutoff is provided to the function via an options object. This way, two different users of this `areEasingsSimilar` don't need to worry about tweaking `SIMILARITY_CUTOFF` constant and affecting the other.

4/10: Using an options object, is nice as it requires the user-code to explicitly name the argument it's passing in so it's easier to read for others later on.

colelawrence (changes requested):
Added suggestions/changes.

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/CurveEditorPopover.tsx
```diff
@@ -1,138 +1,105 @@
...
+import CurveSegmentEditor from './CurveSegmentEditor'
+import EasingOption from './EasingOption'
+import type {CubicBezierPoints} from './shared'
+import {
+  stringFromBezierPoints,
+  bezierPointsFromString,
+  EASING_PRESETS,
+  areEasingsSimilar,
+} from './shared'
+import {COLOR_BASE, COLOR_FOCUS_OUTLINE, COLOR_POPOVER_BACK} from './colors'
```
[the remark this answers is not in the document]
colelawrence: This is not so important to change per se. Just something that comes to mind

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/CurveEditorPopover.tsx
```diff
@@ -233,225 +183,128 @@ const CurveEditorPopover: React.FC<
...
+
+  useEffect(() => {
+    if (!r.current) {
+      const easing: [number, number, number, number] = [
+        trackData.keyframes[index].handles[2],
+        trackData.keyframes[index].handles[3],
+        trackData.keyframes[index + 1].handles[0],
+        trackData.keyframes[index + 1].handles[1],
+      ]
+      setFilter(stringFromBezierPoints(easing))
```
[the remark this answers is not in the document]
colelawrence: For future reference, I seem to have been mistaken on what the values were. the values are actually `[x1, y1, x2, y2]` with the assumption that anchor points are at `0,0` and `1,1`.

Elliot has added a great comment about this in how he's addressed this feedback.

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/CurveEditorPopover.tsx
```diff
@@ -233,225 +183,128 @@ const CurveEditorPopover: React.FC<
...
+    }
+    r.current = false
+  }, [trackData])
+  useLayoutEffect(() => {
+    setTimeout(() => {
+      inputRef.current?.select()
+      inputRef.current?.focus()
+    }, 0)
```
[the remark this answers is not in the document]
colelawrence: Comment now reads `// hack to wait for the paste to actually change the e.target.value` lgtm!

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/EasingOption.tsx
colelawrence: The new comments look great!

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/CurveEditorPopover.tsx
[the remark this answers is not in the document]
thegoleffect: Using a statechart (e.g. xstate) would make it convenient for users to change the bindings, support multiple bindings at once, enable selective binding at the state level, support saving/restoring binding preferences, and for listing what actions are available at any state (and their keybind assignments) [1].

[1] also incredibly helpful for visualizing the available actions, for testing purposes,  etc.

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/colors.ts
thegoleffect: Echoing @colelawrence 's suggestion to use css custom properties, there's new-ish rgb functional notation `--clr-base: 29 29 29;` which can be used like so: `background-color: rgb(var(--clr-base) / 50%)` for dynamic modification.  So you can do alpha math in the previous example but also more complex math on the colors when combined with `calc`. 

Then you only need to define a few theme colors and can generate high contrast or complementary colors on demand from that small starting set (which could be user customizable).

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/EasingOption.tsx
```diff
@@ -0,0 +1,59 @@
...
+`
+
+type IProps = {
+  easing: {
+    label: string
+    value: string
+  }
+} & Parameters<typeof Wrapper>[0]
+
+const EasingOption: React.FC<IProps> = React.forwardRef((props, ref) => {
```
AriaMinaei, line 41: A few visual suggestions:

* Hide the handles from each `EasingOption`?
* A more minimal design for the tooltip.
* Less padding for each `EasingOption`

image

AriaMinaei (commented):
This PR is in a very good state, so I don't have many suggestions. Just left a few comments.

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/CurveEditorPopover.tsx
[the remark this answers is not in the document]
vezwork: Really interesting @thegoleffect. Thanks for the input! Have you seen another library that uses xstate or another statechart to do keybindings?

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/shared.ts
```diff
@@ -0,0 +1,90 @@
...
+  {label: 'Quart Out', value: '0.165, 0.840, 0.440, 1'},
+  {label: 'Quint Out', value: '0.230, 1, 0.320, 1'},
+  {label: 'Expo Out', value: '0.190, 1, 0.220, 1'},
+
+  {label: 'Cubic In Out', value: '0.645, 0.045, 0.355, 1'},
+  {label: 'Quart In Out', value: '0.770, 0, 0.175, 1'},
+  {label: 'Quint In Out', value: '0.860, 0, 0.070, 1'},
+  {label: 'Expo In Out', value: '1, 0, 0, 1'},
+
+  /* These easings are not being included initially in order to
```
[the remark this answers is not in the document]
vezwork: Mariusz and I discussed. My understanding is that we prioritized having three rows with 4 entries each for simplicity. Row 1 includes **linear and sine easings** (most common presets(?)), row 2 includes just **in easings** for 4 different families (most common families(?)), row 3 includes **out easings**, row 4 includes **in-out easings**.

The simplest change I could make to back/circ/quad is to allocate 2 rows for in easings, 2 rows for out easings, and 2 rows for in-out easings

theatre/studio/src/UIRoot/PointerCapturing.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+/**
+ * Introduced `PointerCapturing` for addressing issues with over-shooting easing curves closing the popup preset modal.
+ *
+ * Goal is to be able to determine if the pointer is being captured somewhere in studio (e.g. dragging).
+ *
+ * Some other ideas we considered before going with the PointerCapturing provider and context
+ * - provider: `onPointerCaptureChanged`
+ * - `onDragging={isMouseActive = true}` / `onMouseActive={isMouseActive = true}`
+ * - dragging tracked application wide (ahistoric state) in popover
```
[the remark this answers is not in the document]
vezwork: changed comment to say ephemeral

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/CurveSegmentEditor.tsx
```diff
@@ -0,0 +1,288 @@
...
+          tempTransaction = undefined
+
+          tempTransaction = getStudio()!.tempTransaction(({stateEditors}) => {
+            stateEditors.coreByProject.historic.sheetsById.sequence.replaceKeyframes(
+              {
+                ...props.leaf.sheetObject.address,
+                snappingFunction: val(props.layoutP.sheet).getSequence()
+                  .closestGridPosition,
+                trackId: props.leaf.trackId,
+                keyframes: [setKeyframe(dx, dy)],
```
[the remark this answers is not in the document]
vezwork: This is very tricky. I tried and failed. Background: our dragged position is currently calculated using the control handle positions at the start of the drag and the drag dx and dy. Here is the approach I took:

**Scaling the vertical drag amount by the current height of the unit space.** This fails because the curve height scales the vertical drag, which in turn scales the curve height, and on and on in a loop. It results in compounding and massive vertical drag movements when dragging out of the [0,1] vertical space. AND even when dragging in to the space, the cursor gets offset from the handle as the extremum space changes. This approach does fix it so that once you get back to `extremum space = unit space` then there is no drift, but at the expense that your cursor is now offsets.

By the way I am currently thinking about this, in order to mitigate the compounding effect and mitigate the cursor drift as the extremum space is changing, I think you would either need to do some calculus or think about the dragging in a different way.

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/shared.ts
vezwork: Done

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/shared.ts
vezwork: Done

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/colors.ts
vezwork: These are really interesting and useful points and techiques @colelawrence and @thegoleffect. I am going to take note of them. I think it is out of scope for this PR but I want to come back to this.

vezwork (commented):
This PR in ready for merge on my side.

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/shared.ts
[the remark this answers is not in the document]
AriaMinaei: Can check with users, but I'm 6/10 confident [all](https://github.com/theatre-js/theatre/issues/28#issuecomment-938752916) the presets need to be there.

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/CurveSegmentEditor.tsx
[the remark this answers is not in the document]
AriaMinaei: Okay I think I missed that the drift only occurs with overshoots/space-bending, so I think it's fine. I'll mark this as resolved. Thanks for the explanation.

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/CurveEditorPopover.tsx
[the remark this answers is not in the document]
colelawrence: @thegoleffect ditto. I haven't considered state charts for managing this kind of business logic, but it sounds really interesting, esp in the context of testability. I'll look into it as well.

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/shared.ts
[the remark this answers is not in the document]
vezwork: Done. There are 3 columns: **out**, **in out**, and **in** now

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/CurveEditorPopover.tsx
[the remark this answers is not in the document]
thegoleffect: @vezwork Almost all of the OSS libraries I've seen for this have drawbacks (non-configurable debounce, some teardown & setup ALL keybinds whenever you change one, or they won't integrate well with any state mgmt library). Also, if one were to be built with xstate, it wouldn't necessarily integrate with your workflow - you might want to use actors, web workers / threads, parallel machines, etc. which all require different integration. Thankfully, it is easy enough to set up on your own. 

For vidbase, I initially integrated [react-hotkeys-hook](https://www.npmjs.com/package/react-hotkeys-hook) (and tried a few others) with our xstate machines but ended up just managing the keybinds ourselves for max performance, max compatibility, and better dx. I found out later that linear.app's code shows they do the keybindings themselves as well hence the super high performance there.

It is extra nice b/c xstate has a method you can use to pull out all the available actions one can make for a given state and we used that to auto generate our command palette actions. We can also store state history for playback & debugging. There are lot of free wins.

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/Connector.tsx
```diff
@@ -64,30 +72,40 @@ const Container = styled.div<{isSelected: boolean}>`
...
 
-  const connectorLengthInUnitSpace = next.position - cur.position
-
   const [nodeRef, node] = useRefAndState<HTMLDivElement | null>(null)
 
+  const {isPointerBeingCaptured} = usePointerCapturing(
+    'KeyframeEditor Connector',
+  )
+
   const [popoverNode, openPopover, closePopover, isPopoverOpen] = usePopover(
```
AriaMinaei, line 100: Bug here:

https://user-images.githubusercontent.com/593118/166145346-2a143cf9-e34e-4e62-a98e-7b0d750f2725.mp4

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/CurveEditorPopover.tsx
```diff
@@ -143,315 +124,285 @@ const CurveEditorPopover: React.FC<
...
+      <OptionsContainer ref={optionsContainerRef}>
+        {displayedPresets.map((preset, index) => (
+          <EasingOption
+            key={preset.label}
+            easing={preset}
+            tabIndex={0}
+            onKeyDown={onEasingOptionKeydown(preset)}
+            ref={optionsRef.current[preset.label]}
+            onMouseOver={() => {
+              isFilterSetBeingUsed.current = true
```
AriaMinaei, line 360: Not sure about the UX here. A the user, I'd probably expect the `EasingOption` to only be active while the mouse is hovering it. Once the mouse leaves, it should revert back to whatever was selected before the hover.

https://user-images.githubusercontent.com/593118/166145561-db32dac7-919d-40c8-a3b5-d2522d351f7e.mp4

AriaMinaei (changes requested):
Found a couple of rather small bugs, but other than that, good to merge!

vezwork:
> Found a couple of rather small bugs, but other than that, good to merge!

https://user-images.githubusercontent.com/11082236/166231966-3d07be98-9459-4914-8767-e881bb971fb7.mp4

Fixed the bugs. Fix demo here ^

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/Connector.tsx
vezwork: Fixed. Constrained the curve editor popover to within the horizontal space of the right part of the dope sheet. Added `minX` and `maxX` options to `usePopover` then set these values based on `layoutP.rightDims`.

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/CurveEditorPopover.tsx
vezwork: Fixed. Made it so hovering only changes the curve while you are hovering. If you made edits to the curve before hovering, those edits are still present after you stop hovering. This was achieved by adding a second temporary transaction `hoverTransaction` alongside the existing `editTransaction` (previously called `tempTransaction`)

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/Connector.tsx
vezwork: See my latest comment on the PR for a video.

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/CurveEditorPopover.tsx
vezwork: See my latest comment on the PR for a video.

AriaMinaei (approved):

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/SVGCurveSegment.tsx
vezwork: Ah nvm, I see that this is being reused. I understand the value in this. Just had to consider it a bit more.

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/CurveEditorPopover.tsx
```diff
@@ -287,12 +295,12 @@ const CurveEditorPopover: React.FC<IProps> = (props) => {
     trackData.keyframes[index + 1].handles[1],
   ]
 
-  const isFilterSetBeingUsed = useRef(false)
+  const isFilterUpdatePrevented = useRef(false)
```
colelawrence, line 298: Change: Add docs describing the UX you're going for, here. What kinds of filter update cases are you trying to prevent? What are some weird cases you're avoiding w.r.t. filtering / searching?

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/CurveEditorPopover.tsx
```diff
@@ -352,9 +359,14 @@ const CurveEditorPopover: React.FC<IProps> = (props) => {
...
+            onMouseOver={() => {
+              isFilterUpdatePrevented.current = true
+              editorState.hoverValue(preset.value)
+            }}
+            onMouseOut={() => editorState.discardHoverValue()}
             onClick={() => {
               setHighlightedIndex(displayedPresets.indexOf(preset))
-              editorState.temporarilySetValue(preset.value)
+              editorState.setValue(preset.value)
               //props.onRequestClose()
```
colelawrence, line 370: Change: Delete commented code or document it's purpose.

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/CurveEditorPopover.tsx
```diff
@@ -370,6 +382,9 @@ const CurveEditorPopover: React.FC<IProps> = (props) => {
             )}
           />
         ))}
+        {displayedPresets.length === 0 ? (
+          <NoResultsFoundContainer>No results found</NoResultsFoundContainer>
+        ) : undefined}
```
colelawrence, line 387: For @AriaMinaei

Consider the following just in case there is some oddity with typing in something someone expects to work, but it doesn't for whatever reason.
```suggestion
        {displayedPresets.length === 0 ? (
          <NoResultsFoundContainer>No results found for "{filter}".</NoResultsFoundContainer>
        ) : undefined}
```

Also consider when there are no results, to also follow the "No results found" with the list of all the other results.

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/CurveSegmentEditor.tsx
```diff
@@ -16,40 +17,43 @@ const PATTERN_DOT_SIZE = 0.01
...
-  r: 0.1px;
+  r: 0.09px;
   cursor: move;
   ${pointerEventsAutoInNormalMode};
   &:hover {
+    opacity: 0.4;
   }
   &:hover + ${Circle} {
-    r: 0.1px;
+    fill: #67dfd8;
```
colelawrence, line 49: Nit: Awkward constant color defined separate from all the other color constants.

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/EasingOption.tsx
```diff
@@ -59,8 +58,8 @@ type IProps = {
 
 const EasingOption: React.FC<IProps> = React.forwardRef((props, ref) => {
   const [tooltip, tooltipHostRef] = useTooltip(
-    {enabled: true, verticalPlacement: props.tooltipPlacement},
-    () => <EasingTooltip>{props.easing.label}</EasingTooltip>,
+    {enabled: true, verticalPlacement: props.tooltipPlacement, verticalGap: 0},
+    () => <EasingTooltip showArrow={false}>{props.easing.label}</EasingTooltip>,
```
colelawrence, line 62: Suggestion: Maybe rename `showArrow` to `showPopoverEdgeTriangle={false}`. Generally, just confused by whether there were arrows within the easing presets preview SVG.

Additionally, apply this rename to `popover`.

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/shared.ts
```diff
@@ -1,7 +1,7 @@
 /**
  * This 4-tuple defines the start control point x1,y1 and the end control point x2,y2
  * of a cubic bezier curve. It is assumed that the start of the curve is fixed at 0,0
- * and the end is fixed at 1,1. x1 must always be >= 0 and x2 must always be <= 1.
+ * and the end is fixed at 1,1. x1 must always be \>= 0 and x2 must always be \<= 1.
```
colelawrence, line 4: Nit
```suggestion
 * and the end is fixed at 1,1. X values must be constrained to `x1 >= 0` and `x2 <= 1`.
```

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/shared.ts
```diff
@@ -37,27 +37,43 @@ export function handlesFromCssCubicBezierArgs(
 /**
  * A collection of cubic-bezier approximations of common easing functions
  * ref: https://developer.mozilla.org/en-US/docs/Web/CSS/easing-function
+ * ref: https://github.com/theatre-js/theatre/issues/28#issuecomment-938752916
```
colelawrence, line 40:
```suggestion
 * ref [GitHub issue 28 comment "michaeltheory's suggested default easing presets"](https://github.com/theatre-js/theatre/issues/28#issuecomment-938752916)
```

theatre/studio/src/uiComponents/Popover/TooltipWrapper.tsx
```diff
@@ -97,7 +103,16 @@ const TooltipWrapper: React.FC<{
...
-    const pos = {left, top}
+    const {
+      minX = -Infinity,
+      maxX = Infinity,
+      minY = -Infinity,
+      maxY = Infinity,
+    } = props
+    const pos = {
+      left: clamp(left, minX, maxX - containerRect.width),
+      top: clamp(top, minY, maxY - -containerRect.height),
```
colelawrence, line 114: Nit:
```suggestion
      top: clamp(top, minY, maxY + containerRect.height),
```

theatre/studio/src/uiComponents/Popover/usePopover.tsx
```diff
@@ -21,6 +21,10 @@ export default function usePopover(
     closeWhenPointerIsDistant?: boolean
     pointerDistanceThreshold?: number
     closeOnClickOutside?: boolean
+    minX?: number
+    maxX?: number
+    minY?: number
+    maxY?: number
```
colelawrence, line 27: Suggestion: Might consider pulling these out into their own singular type alias which can be documented separately. e.g. `type AbsolutePlacementBoxConstraints = { minX?: number, ... }`, then that type can be reused in some of the interesting places.

colelawrence (changes requested):

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/CurveEditorPopover.tsx
AriaMinaei: Playing it out in my head, I think `"no results found for ${filter}"` would help. @vezwork @mariusz-wnek-theatre please try that.

Not sure about showing a list of all other results. It's your call @mariusz-wnek-theatre @vezwork

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/CurveEditorPopover.tsx
```diff
@@ -132,62 +137,66 @@ const CurveEditorPopover: React.FC<IProps> = (props) => {
...
+
+          const args = handlesFromCssCubicBezierArgs(newCurve)
+          if (args === null) return
+
+          editTransaction.current = transactionSetCubicBezier(
+            props,
+            cur,
+            next,
+            args,
+          )
```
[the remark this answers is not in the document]
AriaMinaei: Yeah this is important. `tempTransaction`s are applied on top of each other. I'd suggest using a single `tempTransaction` for the whole popover and handle the override logic internally.

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/CurveEditorPopover.tsx
colelawrence: Resolved to not take any action on the `No results found for "{...}"` here, as it initially feels very redundant.
image

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/CurveEditorPopover.tsx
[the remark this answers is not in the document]
vezwork: That is a good question. In my testing, you are right, the order matters BUT the ordered behaviour matches what I would expect to see as a user. Here's the two notable cases: If I hover, then the hover transaction is newest so the hovered curve shown in the UI; If I edit the curve while hovering (by typing in numbers) then the edit transaction is newest and it is shown in the UI.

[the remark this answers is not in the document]
vezwork: Ah, I didn't see Aria's comment when I commented. I humbly disagree with Aria.

If we want to be able to preview presets via hover AND we want to be able to restore the editor state to how it was before the user hovered over a preset (the so-called desired behaviour), then using two temp transactions is simplest and does not behave strangely. If we don't care about the users edits being destroyed while they hover preview a preset, then using one preset would be fine. It is possible to use a single temp transaction to manage the desired behaviour, but then I will have to introduce additional state and state management logic that would, as far as I can see, behave the same as two temp transactions anyway.

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/CurveEditorPopover.tsx
```diff
@@ -132,62 +137,66 @@ const CurveEditorPopover: React.FC<IProps> = (props) => {
...
+        discardValue(): void {
+          editTransaction.current?.discard()
+          editTransaction.current = undefined
+        },
+        hoverValue(newCurve: string): void {
+          hoverTransaction.current?.discard()
+          hoverTransaction.current = undefined
 
           const args = handlesFromCssCubicBezierArgs(newCurve)
           if (args === null) return
```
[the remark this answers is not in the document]
vezwork: rename `args` -> `handles` taken. 

rename `CssCubicBezierArgs` to `CssCubicBezier` not taken, explanation: A full CSS cubic bezier string looks like
`cubic-bezier(0, 0, 1, 1)`. the "args" part of the name refers specifically to the comma separated substring inside the parentheses of the CSS cubic bezier string i.e. `0, 0, 1, 1`.

Added this comment to above the function definition.

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/Connector.tsx
```diff
@@ -22,29 +22,38 @@ import selectedKeyframeIdsIfInSingleTrack from '@theatre/studio/panels/SequenceE
...
-    return lighten(0.2, CONNECTOR_THEME.normalColor)
+  selectedColor(isPopoverOpen: boolean) {
+    return lighten(0.2, CONNECTOR_THEME.baseColor(isPopoverOpen))
   },
-  get selectedHoverColor() {
-    return lighten(0.4, CONNECTOR_THEME.normalColor)
+  selectedHoverColor(isPopoverOpen: boolean) {
+    return lighten(0.4, CONNECTOR_THEME.baseColor(isPopoverOpen))
   },
 }
```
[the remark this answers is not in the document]
vezwork: We pair programmed this and I like the solution.

```typescript
type IConnectorThemeValues = {
  isPopoverOpen: boolean
  isSelected: boolean
}

export const CONNECTOR_THEME = {
  normalColor: `#365b59`, // (greenish-blueish)ish
  popoverOpenColor: `#817720`, // orangey yellowish
  barColor: (values: IConnectorThemeValues) => {
    const base = values.isPopoverOpen
      ? CONNECTOR_THEME.popoverOpenColor
      : CONNECTOR_THEME.normalColor
    return values.isSelected ? lighten(0.2, base) : base
  },
// [...]
}

const Container = styled.div<IConnectorThemeValues>`
  position: absolute;
  background: ${CONNECTOR_THEME.barColor};
// [...]
```

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/CurveEditorPopover.tsx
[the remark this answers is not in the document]
vezwork: Cole and I talked on this and agreed that the pair of temp transaction maps nicely to the intended behaviour and simplifies the code.

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/CurveEditorPopover.tsx
```diff
@@ -380,6 +395,35 @@ const CurveEditorPopover: React.FC<IProps> = (props) => {
...
+          handles: [cur.handles[0], cur.handles[1], args[0], args[1]],
+        },
+        {
+          ...next,
+          handles: [args[2], args[3], next.handles[2], next.handles[3]],
+        },
+      ],
+    })
+  })
+}
```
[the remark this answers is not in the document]
vezwork: We pair programmed this and pulled out a `useUIOptionGrid` hook.

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/CurveEditorPopover.tsx
vezwork: Deleted

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/EasingOption.tsx
vezwork: renamed in `BasicPopover`, `usePopover` does not use this name so I am not sure what you mean by your second paragraphg.

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/shared.ts
vezwork: Thanks, this was actually incomplete. They should both be in the 0, 1 interval

theatre/studio/src/uiComponents/Popover/TooltipWrapper.tsx
vezwork: lol

theatre/studio/src/uiComponents/Popover/useTooltip.tsx
```diff
@@ -15,6 +15,7 @@ export default function useTooltip(
     enterDelay?: number
     exitDelay?: number
     verticalPlacement?: 'top' | 'bottom' | 'overlay'
+    verticalGap?: number
```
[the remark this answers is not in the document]
vezwork: It has no effect in that case. Added a quick comment.

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/CurveEditorPopover.tsx
```diff
@@ -295,6 +246,15 @@ const CurveEditorPopover: React.FC<IProps> = (props) => {
     trackData.keyframes[index + 1].handles[1],
   ]
 
+  /**
+   * `isFilterUpdatePrevented` is only read in the `useEffect` immediately below.
```
colelawrence, line 250: Praise: Lovely doc

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/CurveEditorPopover.tsx
```diff
@@ -395,13 +394,21 @@ const CurveEditorPopover: React.FC<IProps> = (props) => {
 
 export default CurveEditorPopover
 
+// type CurveController = {
```
colelawrence, line 397: Change: Remove commented code or document the idea it relates to

colelawrence (approved):
One minor comment. LGTM

AriaMinaei:
Found a bug:
Screenflick Movie 86

AriaMinaei (changes requested):
Blocked by this bug:

Screenflick Movie 86

I think the transactions of filters/hovers need to be handled in a more declarative way. Have an idea to share about this in chat

AriaMinaei (changes requested):
Quick review: The `Esc` key doesn't work anymore.

Will do a thorough one in a bit.

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/CurveSegmentEditor.tsx
```diff
@@ -0,0 +1,288 @@
...
+const PATTERN_DOT_SIZE = 0.01
+const PATTERN_DOT_COUNT = 8
+const PATTERN_GRID_SIZE = (1 - PATTERN_DOT_SIZE) / (PATTERN_DOT_COUNT - 1)
+
+const CURVE_START_OVERSHOOT_COLOR = '#3e9bb8'
+const CURVE_START_COLOR = '#40AAA4'
+const CURVE_MID_START_COLOR = '#58C0BA'
+const CURVE_MID_END_COLOR = '#FCAB54'
+const CURVE_END_COLOR = '#F78104'
+const CURVE_END_OVERSHOOT_COLOR = '#f76904'
```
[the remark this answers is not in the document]
vezwork: Added comment

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/CurveSegmentEditor.tsx
```diff
@@ -0,0 +1,288 @@
...
+`
+
+const HitZone = styled.circle`
+  stroke-width: 0.1px;
+  vector-effect: non-scaling-stroke;
+  r: 0.1px;
+  cursor: move;
+  ${pointerEventsAutoInNormalMode};
+  &:hover {
+  }
```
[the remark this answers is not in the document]
vezwork: Removed

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/CurveEditorPopover.tsx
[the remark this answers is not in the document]
vezwork: After talking with Aria a bit more I refactored this to use a single temp transaction.

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/CurveEditorPopover.tsx
```diff
@@ -132,62 +137,66 @@ const CurveEditorPopover: React.FC<IProps> = (props) => {
...
+            args,
+          )
+        },
+        discardValue(): void {
+          editTransaction.current?.discard()
+          editTransaction.current = undefined
+        },
+        hoverValue(newCurve: string): void {
+          hoverTransaction.current?.discard()
+          hoverTransaction.current = undefined
```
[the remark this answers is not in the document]
vezwork: Outdated after refactor

theatre/studio/src/uiComponents/Popover/usePopover.tsx
vezwork: Pulled them out in `usePopover` and `tooltipWrapper`

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/CurveSegmentEditor.tsx
```diff
@@ -0,0 +1,288 @@
...
+        strokeWidth="0.01"
+      />
+      <line
+        x1={0}
+        y1={toExtremumSpace(1)}
+        x2={cur.handles[2]}
+        y2={toExtremumSpace(1 - cur.handles[3])}
+        stroke={CONTROL_COLOR}
+        strokeWidth="0.01"
+      />
```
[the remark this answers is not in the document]
vezwork: Added some comments describing each part

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/CurveSegmentEditor.tsx
```diff
@@ -0,0 +1,288 @@
...
+  setKeyframe: (dx: number, dy: number) => Keyframe,
+): void {
+  const handlers = useFreezableMemo<Parameters<typeof useDrag>[1]>(
+    (setFrozen) => {
+      // Considered using "scrub" instead to manage a not-necessarilly commital change over time
+      // But it appears that scrub doesn't allow to change a "pointer" for the keyframes
+      let tempTransaction: CommitOrDiscard | undefined
+
+      return {
+        debugName: 'CurveSegmentEditor/useRightDrag',
```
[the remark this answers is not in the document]
vezwork: Done

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/SVGCurveSegment.tsx
```diff
@@ -0,0 +1,96 @@
...
+  easing: CubicBezierPoints | null
+}
+
+const SVGCurveSegment: React.FC<IProps> = (props) => {
+  const {easing} = props
+
+  if (!easing) return <></>
+
+  const LEFT_CONTROL_POINT = [easing[0], toVerticalSVGSpace(easing[1])]
+  const RIGHT_CONTROL_POINT = [easing[2], toVerticalSVGSpace(easing[3])]
```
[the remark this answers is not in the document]
vezwork: No good reason! Fixed

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/CurveEditorPopover.tsx
vezwork: Outdated after refactor

AriaMinaei (changes requested):
Still can't cancel changes by hitting `Esc`.

Hitting `Esc` should close the popover and return the tween to its original value before the popover was open.

Screenflick Movie 87

AriaMinaei (approved):
Never mind. It was my Vim bindings preventing "Esc"

PR #123, opened 2022-04-13
Bump 'react-shadow' to v19.0.3

donmccurdy:
Picking up a recent fix from https://github.com/Wildhoney/ReactShadow/pull/121.

Without this, we needed to include a hack (below) in our code, so this just enables a bit of cleanup.

```html
window.global = window;
```

### DIFF
--- a/packages/r3f/package.json
+++ b/packages/r3f/package.json
@@ -48,7 +48,7 @@
     "polished": "^4.1.3",
     "react-icons": "^4.2.0",
     "react-merge-refs": "^1.1.0",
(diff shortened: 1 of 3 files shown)


vezwork:
@AndrewPrifer are you able to verify that this change is okay?

AndrewPrifer:
@vezwork Should not affect theatre in any way, this fix is only because reactshadow, the library we use for shadow dom was previously relying on the non-standard global keyword which is transformed by webpack but not by other build tools. I think this is safe to merge as is.

PR #127, opened 2022-04-19
Dev/cl 2022 04/refactors docs

colelawrence:
General documentation, code cleanup, and similar that has fallen out of work to understand approaches for sequence markers.

### DIFF
--- a/theatre/core/src/propTypes/internals.ts
+++ b/theatre/core/src/propTypes/internals.ts
@@ -17,8 +17,16 @@ export type IValidCompoundProps = {
   [K in string]: PropTypeConfig
 }
 
-// Question: Why doesn't this include Rgba? Should it include Marker?
-//  - Is this just for compound props?
+/**
+ *
+ * This does not include Rgba since Rgba does not have a predictable
+ * object shape. We prefer to infer that compound props are described as
+ * `Record<string, IShorthandProp>` for now.
+ *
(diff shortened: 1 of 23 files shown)


theatre/core/src/propTypes/internals.ts
```diff
@@ -17,6 +17,8 @@ export type IValidCompoundProps = {
   [K in string]: PropTypeConfig
 }
 
+// Question: Why doesn't this include Rgba? Should it include Marker?
+//  - Is this just for compound props?
```
AriaMinaei, line 21: Because this line would become unpredictable:

```typescript
const obj = sheet.object('obj', {color: {r: 0, b: 0, g: 0, a: 0}})
```

* In Theatre 0.4.7, `obj.value.color` would be a compound prop. It cannot be sequenced, but its sub-props can.
* In Theatre 0.4.8, `obj.value.color` would be a simple prop. It _can_ be sequenced, while it sub-props _can't_.

AriaMinaei:
@colelawrence Okay, added an answer to the one question that wasn't part of the previous review.

theatre/core/src/propTypes/internals.ts
colelawrence: I've applied a comment to this effect, I think.

vezwork (approved):

PR #137, opened 2022-04-28
Fix typo?

vezwork:
On dataverse naming convention docs. It says that derivations have the convention `variableP` but I think it should say `variableD`?


fulopkovacs (approved):
LGTM :relaxed:

PR #139, opened 2022-05-02
Make editable schema-based

AndrewPrifer:
This PR is centered around one main issue: all objects and their properties are hard-coded in the r3f extension right now. This means that there is not only no possiblity for user-land extension to the supported object types, but it is also hard for us to change or add things to it.

This PR solves it by creating and editable factory function that takes a schema and creates the editable component based on that schema.

This way all the things necessary for supporting an object (properties, how they translate to theatre, whether they have transform controls, what icons to use for them in the side bar, what helper to display in the viewport, etc) can be defined in a single place with a few lines of code.

---

This PR ended up doing a bunch of things. First off, the schema definition.

The schema is of shape
```ts
export type Helper = Object3D & {
  update?: () => void
}
type PropConfig<T> = {
  parse: (props: Record<string, any>) => T
  apply: (value: T, object: any) => void
  type: IShorthandCompoundProps
}
type Props = Record<string, PropConfig<any>>
type Meta<T> = {
  useTransformControls: boolean
  updateObject?: (object: T) => void
  icon: IconID
  dimensionless?: boolean
  createHelper: (object: T) => Helper
}
export type ObjectConfig<T> = {props: Props} & Meta<T>
export type EditableFactoryConfig = Partial<
  Record<keyof JSX.IntrinsicElements, ObjectConfig<any>>
>
```

and lets us define all that is needed for the r3f extension to support a certain set of objects.

Prop configs let us support arbitrary properties that are defined through an arbitrary set of props passed to the react component. It does this through
- the `parse()` function that receives all the props passed, and returns the initial value of the theatre prop
- the `apply()` function that applies values received from theatre to the object

Second, the PR improves the UX of the viewport. Helpers are no longer displayed for dimensionless objects all the time, cluttering the viewport, only for hovered and selected objects. Selection colors are also aligned with theatre designs.

Third, it expands the number of supported objects with points, lines, line segments and line loops and makes the most common properties of existing objects editable, e.g. angle, penumbra and intensity for spotlights.

In the future the schema API and the editable factory will be exposed to users so anyone can extend the supported objects if the default set doesn't cover their needs.

### DIFF
--- a/packages/playground/src/shared/r3f-rocket/App.tsx
+++ b/packages/playground/src/shared/r3f-rocket/App.tsx
@@ -1,5 +1,5 @@
 import {editable as e, RefreshSnapshot, SheetProvider} from '@theatre/r3f'
-import {OrbitControls, Stars} from '@react-three/drei'
+import { Stars} from '@react-three/drei'
 import {getProject} from '@theatre/core'
 import React, {Suspense, useState} from 'react'
 import {Canvas} from '@react-three/fiber'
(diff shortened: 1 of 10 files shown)


AriaMinaei (approved):
LGTM!

PR #140, opened 2022-05-03
Markers

vezwork:
## Description

Markers are small indicators that can be added to a timeline. They simply mark a certain time. They could be used to denote a "change of action" or the start or end of a new part of the animation. They can be dragged, snapped, and added/deleted, but not much else for now.

## Demo
Ready for handoff to Aria and Mariusz for design finalization: 
https://user-images.githubusercontent.com/11082236/167178008-29b4e804-70f5-4644-ac2c-d2b1832960c0.mp4


theatre/studio/src/uiComponents/PointerEventsHandler.tsx
```diff
@@ -94,6 +96,8 @@ const PointerEventsHandler: React.FC<{
  * but then "unlocking" that style will again reveal the existing styles.
  *
  * It behaves a bit like a stack.
+ *
+ * @see lockedCursorCssPropName
```
colelawrence, line 100: Question: Does this creates a link just like `{@link lockedCursorCssPropName}`?

theatre/studio/src/uiComponents/PointerEventsHandler.tsx
```diff
@@ -16,6 +16,8 @@ const elementId = 'pointer-root'
  * whose value will be the locked cursor (e.g. ew-resize).
  *
  * Look up references of this constant for examples of how it is used.
+ *
+ * @see useCssCursorLock - code that locks the cursor
  */
 export const lockedCursorCssPropName = '--lockedCursor'
```
colelawrence, line 22: Nit: `lockedCursorCssPropName` is more like `lockedCursorCssVarName`, right?

theatre/studio/src/store/types/historic.ts
```diff
@@ -71,6 +71,9 @@ export type StudioHistoricState = {
           {
             selectedInstanceId: undefined | string
             sequenceEditor: {
+              markers?: {
+                time: number
+              }[]
```
colelawrence, line 76: Question: Should we ad docs to say what unit `time` is in? Is it in seconds?

colelawrence, line 76: Change: It seems like markers should individually have their own identifiers as well (something as simple as `Date.now().toString(36)` would be fine to generate an id in mvp). This would ensure we can talk about markers based on their `id` rather than just the time they sit on (preventing multiple markers from being in the same place).

colelawrence: Side note: This could be fun to use `Nominal`s on after my other refactor eventually lands.

theatre/studio/src/store/stateEditors.ts
```diff
@@ -245,6 +245,42 @@ namespace stateEditors {
...
+                    marker,
+                  ]
+                }
+              }
+
+              export function removeMarker(p: SheetAddress, atTime: number) {
+                const sequenceEditor =
+                  stateEditors.studio.historic.projects.stateByProjectId.stateBySheetId._ensure(
+                    p,
+                  ).sequenceEditor
```
colelawrence, line 274: Nit: `p` is kinda weird shortened identifier

theatre/studio/src/panels/SequenceEditorPanel/RightOverlay/Playhead.tsx
```diff
@@ -305,3 +314,27 @@ const Playhead: React.FC<{layoutP: Pointer<SequenceEditorPanelLayout>}> = ({
...
+  layout: SequenceEditorPanelLayout,
+  time: number,
+) {
+  return useContextMenu(node, {
+    items: () => {
+      return [
+        {
+          label: 'Place marker',
+          callback: () => {
+            getStudio()!.transaction(({stateEditors}) => {
```
colelawrence, line 329: Nit: I don't think the `!` on `getStudio()!` is necessary

theatre/studio/src/panels/SequenceEditorPanel/RightOverlay/Playhead.tsx
```diff
@@ -305,3 +314,27 @@ const Playhead: React.FC<{layoutP: Pointer<SequenceEditorPanelLayout>}> = ({
 }
 
 export default Playhead
+
+function useMarkerContextMenu(
+  node: HTMLElement | null,
+  layout: SequenceEditorPanelLayout,
+  time: number,
+) {
```
colelawrence, line 322: Suggestion: `useMarkerContextMenu` -> `usePlayheadContextMenu`

theatre/studio/src/panels/SequenceEditorPanel/RightOverlay/Marker.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+  const [contextMenu] = useMarkerContextMenu(node, layout, marker.time)
+  return (
+    <>
+      {contextMenu}
+      <Mark
+        ref={ref}
+        {...{[attributeNameThatLocksFramestamp]: marker.time.toFixed(3)}}
+        data-pos={marker.time.toFixed(3)}
+        left={clippedSpaceFromUnitSpace(marker.time)}
```
colelawrence, line 53: Nit: Just from reading this, I'm not exactly sure what the complexity was for these props around here.
Maybe a couple comments with a small amount of clarification could help. Something short even like `data-pos={marker.time.toFixed(3)} // used for snapping target` (I actually don't know what `data-pos` is for, just an example of being able to keep it short).

theatre/studio/src/panels/SequenceEditorPanel/RightOverlay/Marker.tsx
```diff
@@ -0,0 +1,83 @@
...
+  overflow: hidden;
+  font-size: 10px;
+  left: ${({left}) => left}px;
+
+  ${pointerEventsAutoInNormalMode};
+
+  #pointer-root.draggingPositionInSequenceEditor & {
+    pointer-events: auto;
+    cursor: var(${lockedCursorCssPropName});
+  }
```
colelawrence, line 29: Question: What is the `&` for in the selector? I know how it works in `& div` in side a rule body, but not sure what it would do here.

theatre/studio/src/panels/SequenceEditorPanel/RightOverlay/Marker.tsx
colelawrence, line 53: Hmm: The way this "left" is passed down feels a bit itchy to me. It literally means left with absolute for the styling. So, maybe just renaming it to `cssLeft` would help at least keep that awkwardness exposed a bit better in case there's something more interesting later (like a time -> left alignment css provider hook or something later, i dunno).

theatre/studio/src/panels/SequenceEditorPanel/RightOverlay/Marker.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+  &:hover {
+    background-color: red;
+  }
+`
+
+const Marker: React.VFC<{
```
colelawrence, line 36: Praise: Nice VFC 😄

theatre/studio/src/panels/SequenceEditorPanel/RightOverlay/Marker.tsx
```diff
@@ -0,0 +1,83 @@
...
+  layoutP: Pointer<SequenceEditorPanelLayout>
+  marker: {time: number}
```
colelawrence, line 38: Hmm: A lot of `{time: number}` repeated types, but it doesn't really bother me, since I don't necessarily want to commit to which models are actually shared for these signatures.

theatre/studio/src/panels/SequenceEditorPanel/RightOverlay/Marker.tsx
```diff
@@ -0,0 +1,83 @@
...
+import getStudio from '@theatre/studio/getStudio'
+import {lockedCursorCssPropName} from '@theatre/studio/uiComponents/PointerEventsHandler'
+import useContextMenu from '@theatre/studio/uiComponents/simpleContextMenu/useContextMenu'
+import useRefAndState from '@theatre/studio/utils/useRefAndState'
+import React from 'react'
+import styled from 'styled-components'
+import {attributeNameThatLocksFramestamp} from '@theatre/studio/panels/SequenceEditorPanel/FrameStampPositionProvider'
+import type {SequenceEditorPanelLayout} from '@theatre/studio/panels/SequenceEditorPanel/layout/layout'
+
+const Mark = styled.div<{left: number}>`
```
colelawrence, line 13: Suggestion: Try out a `IMarkThemeValues` type to see what it looks like, maybe? I don't feel strongly, but as you add colors it could make more sense.

colelawrence (commented):
Initial feelings. Not anything required to apply at this time considering it's a draft.

theatre/studio/src/store/types/historic.ts
AriaMinaei: Yeah some identifier would be necessary to track concurrent changes between multiple users. Have a look at `generateKeyframeId()` for an example.

AriaMinaei: Yeah some identifier would be necessary to track concurrent changes between multiple users. Have a look at `generateKeyframeId()` for an example.

AriaMinaei: I'd say just add an identifier like `{time: number, id: string}`. No need to do `{list: ID[], byId: Record<ID, {time: number}>}`. The latter is more efficient, but there won't be that many markers per sequence anyway.

theatre/studio/src/panels/SequenceEditorPanel/RightOverlay/TopStrip.tsx
```diff
@@ -34,8 +34,9 @@ const TopStrip: React.FC<{layoutP: Pointer<SequenceEditorPanelLayout>}> = ({
 
   return (
     <>
-      <Container {...{[attributeNameThatLocksFramestamp]: 'hide'}}>
+      <Container>
```
AriaMinaei, line 37: why?

theatre/studio/src/panels/SequenceEditorPanel/RightOverlay/Playhead.tsx
```diff
@@ -249,6 +251,12 @@ const Playhead: React.FC<{layoutP: Pointer<SequenceEditorPanelLayout>}> = ({
   // hide the frame stamp when seeking
   useLockFrameStampPosition(useVal(layoutP.seeker.isSeeking) || isDragging, -1)
 
+  const [contextMenu] = useMarkerContextMenu(
+    thumbNode,
+    val(layoutP),
```
AriaMinaei, line 256: Multiple issues here.

1 - `val()` can _only_ be called inside a `prism()`. Otherwise nothing will track the changes to the underlying derivation.
You're calling it inside a react component. Either use `useVal()` or `usePrism()` here.

AriaMinaei: 2 - There would still be a problem if you corrected this to to `usePrism(() => val(layoutP))`.

It would work, but it would be inefficient. Any changes to any part of the layout that is unrelated to markers would still cause the `Playhead` to re-render.

To fix this, you need to point inside `layoutP`, for example: `layoutP.sheet`, which would update only when the user picks a new sheet.

AriaMinaei: 3 - We don't even need to know the `val(layoutP.sheet)` or `val(layoutP.sheet).getSequence().position` unless the user actually R-clicks on the playhead.

So better to move these calculations to the `items()` callback in `useContextMenu(node, {items})`.

There, you _can_ use `val()` which won't be reactive, but you won't need reactivity there.

theatre/studio/src/panels/SequenceEditorPanel/RightOverlay/Markers.tsx
```diff
@@ -0,0 +1,36 @@
...
+  width: 100%;
+  height: 40px;
+  overflow: hidden;
+`
+
+const Markers: React.VFC<{layoutP: Pointer<SequenceEditorPanelLayout>}> = ({
+  layoutP,
+}) => {
+  const sheetAddress = useVal(layoutP).sheet.address
+  const markers = useVal(getStudio()!.atomP)?.historic.projects
```
AriaMinaei, line 22: These two lines are also calling `useVal()` with a close-to-root pointer.

Suggestion:

```
const sheet = useVal(layoutP.sheet)
const markers = useVal(getStudio()!.atomP.historic.projects.stateByProjectId[sheet.address.projectId].stateBySheetId[sheetAddress.sheetId].sequenceEditor.markers)
```

Good thing is, you won't even need to do all the optional chaining (`?.`) because the return type of `useVal()` will be `T | undefined` if there are any `undefined`s in the path.

theatre/studio/src/panels/SequenceEditorPanel/RightOverlay/Marker.tsx
```diff
@@ -0,0 +1,83 @@
...
+  &:hover {
+    background-color: red;
+  }
+`
+
+const Marker: React.VFC<{
+  layoutP: Pointer<SequenceEditorPanelLayout>
+  marker: {time: number}
+}> = ({layoutP, marker}) => {
+  const layout = useVal(layoutP)
```
AriaMinaei, line 40: As another example, every `Marker` element will be recalculated every time anything in `layoutP` changes, like if `val(layoutP).seeker.isSeeking` changes, we'll still recalculate all the `Marker`s.

[ ] Action item: find a place to document this.

theatre/studio/src/panels/SequenceEditorPanel/RightOverlay/Marker.tsx
```diff
@@ -0,0 +1,83 @@
...
+const Mark = styled.div<{left: number}>`
+  position: absolute;
+  height: 20px;
+  width: 8px;
+  background-color: white;
+  z-index: 1;
+  color: black;
+  overflow: hidden;
+  font-size: 10px;
+  left: ${({left}) => left}px;
```
AriaMinaei, line 22: Two issues:

1. The horizontal placement of this element will change every frame, every time the user pans/zooms. So we need to make sure it renders in an optimal way. Changing most css props like `left, top, width, margin, padding`  will cause the browser's rendering engine to re-layout and possibly re-paint this element and its parents/siblings and all their descendants. That's inefficient. Instead, you need to use `transform` so the browser can skip re-layout and re-paint, and likely offload the compositing to the GPU.
2. Changing props that you pass to styled-components will cause styled-components to create a new class name for each new permutation of all props values. This is fine for many props, but `left` will be changing once a frame per marker, so it can't be passed as a styled-component prop. Use `data-pos` or just set the `transform` directly from your react component like `<Mark style={{transform}}>`

AriaMinaei (changes requested):

theatre/studio/src/store/stateEditors.ts
```diff
@@ -245,6 +245,42 @@ namespace stateEditors {
                   ])
                 }
               }
+
+              export function replaceMarker(
+                p: SheetAddress,
+                marker: {time: number},
```
AriaMinaei, line 251: as a convention, all fns in `stateEditors` take a single argument. Reason is, all arguments will have a unique prop name, which should make it a bit easier to later these state editors to a concurrent editing setup.

theatre/studio/src/panels/SequenceEditorPanel/RightOverlay/Marker.tsx
vezwork: Noted and discussed briefly.

theatre/studio/src/store/stateEditors.ts
vezwork: We renamed it to `options`

theatre/studio/src/panels/SequenceEditorPanel/RightOverlay/Marker.tsx
vezwork: Added a comment 

```
// "All instances of this component inside #pointer-root when it has the .draggingPositionInSequenceEditor class"
  // ref: https://styled-components.com/docs/basics#pseudoelements-pseudoselectors-and-nesting
  ```

File: theatre/studio/src/panels/SequenceEditorPanel/RightOverlay/Marker.tsx (hunk 7 above)
Comment (vezwork) [reply theatre/studio/src/panels/SequenceEditorPanel/RightOverlay/Marker.tsx]: Aria had a comment about this below too

File: theatre/studio/src/panels/SequenceEditorPanel/RightOverlay/Marker.tsx (hunk 16 above)
Comment (vezwork) [reply theatre/studio/src/panels/SequenceEditorPanel/RightOverlay/Marker.tsx:23]: Cole had a comment about this above

File: theatre/studio/src/panels/SequenceEditorPanel/RightOverlay/Marker.tsx (hunk 11 above)
Comment (vezwork) [reply theatre/studio/src/panels/SequenceEditorPanel/RightOverlay/Marker.tsx:14]: `left` was refactored out

File: theatre/studio/src/panels/SequenceEditorPanel/RightOverlay/Marker.tsx [hunk 18]
```diff
@@ -37,14 +39,28 @@ const MarkerDot = styled.div`
 
 const Marker: React.VFC<{
   layoutP: Pointer<SequenceEditorPanelLayout>
-  marker: StudioHistoricStateSequenceEditorMarker
-}> = ({layoutP, marker}) => {
-  const layout = useVal(layoutP)
-  const clippedSpaceFromUnitSpace = layout.clippedSpace.fromUnitSpace
+  markerId: SequenceMarkerId
+}> = ({layoutP, markerId}) => {
+  const sheetAddress = val(layoutP.sheet.address)
```
AriaMinaei, line 44: `useVal()`

theatre/studio/src/panels/SequenceEditorPanel/RightOverlay/Marker.tsx
vezwork: Fixed to be more specific.

vezwork (commented):
LGTM. 

For future: awkward use of `Object.assign(...)` in stateEditors

theatre/studio/src/panels/SequenceEditorPanel/RightOverlay/MarkerDot.tsx
```diff
@@ -0,0 +1,305 @@
...
+    layoutP,
+    marker,
+  })
+
+  return (
+    <div
+      style={{
+        position: 'absolute',
+        zIndex: zIndexes.marker,
+        transform: `translateX(${clippedSpaceFromUnitSpace(
```
AriaMinaei, line 149: This would still render when out of bands:

https://user-images.githubusercontent.com/593118/167209870-37bd281f-ef75-467d-a966-c597cf1bc877.mp4

To fix it, check if the marker is withing `[0, val(layoutP.clippedSpace.width)]` and if not, set its `translateX(-1000) scale(0, 0)` so it disappears but still remains in the tree and doesn't cause reflow/repaint.

theatre/studio/src/panels/SequenceEditorPanel/RightOverlay/MarkerDot.tsx
```diff
@@ -0,0 +1,305 @@
...
+          marker.position,
+        )}px)`,
+        // below the sequence ruler "top bar"
+        top: '18px',
```
AriaMinaei, line 153: Better put this in the style rule

AriaMinaei (changes requested):

AriaMinaei (approved):

PR #142, opened 2022-05-03
Add method to clear persistent storage

donmccurdy:
Possible solution to an issue we discussed briefly in Discord — our application registers an extension, and displays a button allowing users to save their changes to a file. We enable persistent storage in case the page crashes or reloads before changes have been saved. In some cases the user may decide _not_ to save the changes they've made since the last save, in which case we need a way to clear persistent storage.

I've tried just clearing all localstorage entries with the right prefix and then reloading, but then theatre.js is doing one last commit to localstorage before the page reloads and the cache persists. This PR would expose a new method, `studio.clearPersistentStorage()`, which could be used by userland code or extensions.

*This contribution is funded by The New York Times.*

### DIFF
--- a/theatre/studio/src/Studio.ts
+++ b/theatre/studio/src/Studio.ts
@@ -519,6 +519,6 @@ export class Studio {
   }
 
   clearPersistentStorage(persistenceKey = DEFAULT_PERSISTENCE_KEY) {
-    this._store.clearPersistentStorage(persistenceKey)
+    this._store.__experimental_clearPersistentStorage(persistenceKey)
   }
 }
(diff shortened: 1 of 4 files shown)


AriaMinaei (approved):

AriaMinaei:
Thanks @donmccurdy! Just note that I've renamed the API to: `studio.__experimental.__experimental_clearPersistentStorage(persistenceKey)`

craigthings:
I wonder if an alternative to this would be to add a config param when initializing studio to disable persistent storage? there are some cases, I imagine, where storing and restoring state is managed in userland.

PR #144, opened 2022-05-04
Chore/cl 2022 05/refactor and docs from inline editing feature

colelawrence:
See commits list in order.

Commits copied here for clarity:

> dev(tsconfig): ensure imports use `import type ...`
 
> refactor: add pointer types for prop editors 
 
> refactor: Add working Nominal types, clarify identifiers
>  * Use more Nominal types to help with internal code id usage consistency
>  * Broke apart StudioHistoricState type

> doc + optimize: prism.ts (use Maps)

### DIFF
--- a/packages/dataverse/src/pointer.ts
+++ b/packages/dataverse/src/pointer.ts
@@ -33,6 +33,11 @@ const cachedSubPathPointersWeakMap = new WeakMap<
  * A wrapper type for the type a `Pointer` points to.
  */
 export type PointerType<O> = {
+  /**
+   * Only accessible via the type system.
+   * This is a helper for getting the underlying pointer type
+   * via the type space.
+   */
   $$__pointer_type: O
 }
 
@@ -53,21 +58,28 @@ export type PointerType<O> = {
  *
  * The current solution is to just avoid using `any` with pointer-related code (or type-test it well).
  * But if you enjoy solving typescript puzzles, consider fixing this :)
- *
+ * Potentially, [TypeScript variance annotations in 4.7+](https://devblogs.microsoft.com/typescript/announcing-typescript-4-7-beta/#optional-variance-annotations-for-type-parameters)
+ * might be able to help us.
  */
 export type Pointer<O> = PointerType<O> &
-  (O extends UnindexableTypesForPointer
-    ? UnindexablePointer
-    : unknown extends O
-    ? UnindexablePointer
-    : O extends (infer T)[]
-    ? Pointer<T>[]
-    : O extends {}
-    ? {
-        [K in keyof O]-?: Pointer<O[K]>
-      } /*&
-        {[K in string | number]: Pointer<K extends keyof O ? O[K] : undefined>}*/
-    : UnindexablePointer)
+  // `Exclude<O, undefined>` will remove `undefined` from the first type
+  // `undefined extends O ? undefined : never` will give us `undefined` if `O` is `... | undefined`
+  PointerInner<Exclude<O, undefined>, undefined extends O ? undefined : never>
+
+// By separating the `O` (non-undefined) from the `undefined` or `never`, we
+// can properly use `O extends ...` to determine the kind of potential value
+// without actually discarding optionality information.
+type PointerInner<O, Optional> = O extends UnindexableTypesForPointer
+  ? UnindexablePointer
+  : unknown extends O
+  ? UnindexablePointer
+  : O extends (infer T)[]
+  ? Pointer<T>[]
+  : O extends {}
+  ? {
+      [K in keyof O]-?: Pointer<O[K] | Optional>
+    }
+  : UnindexablePointer
 
 const pointerMetaSymbol = Symbol('pointerMeta')
 
--- a/theatre/core/src/projects/store/types/SheetState_Historic.ts
+++ b/theatre/core/src/projects/store/types/SheetState_Historic.ts
@@ -1,5 +1,13 @@
-import type {KeyframeId, SequenceTrackId} from '@theatre/shared/utils/ids'
-import type {SerializableMap, StrictRecord} from '@theatre/shared/utils/types'
+import type {
+  KeyframeId,
+  ObjectAddressKey,
+  SequenceTrackId,
+} from '@theatre/shared/utils/ids'
+import type {
+  SerializableMap,
+  SerializableValue,
+  StrictRecord,
+} from '@theatre/shared/utils/types'
 
 export interface SheetState_Historic {
   /**
@@ -10,14 +18,13 @@ export interface SheetState_Historic {
    * of another state, it will be able to inherit the overrides from ancestor states.
    */
   staticOverrides: {
-    byObject: StrictRecord<string, SerializableMap>
+    byObject: StrictRecord<ObjectAddressKey, SerializableMap>
   }
-  sequence?: Sequence
+  sequence?: HistoricPositionalSequence
 }
 
-type Sequence = PositionalSequence
-
-type PositionalSequence = {
+// Question: What is this? The timeline position of a sequence?
+export type HistoricPositionalSequence = {
   type: 'PositionalSequence'
   length: number
   /**
@@ -27,7 +34,7 @@ type PositionalSequence = {
   subUnitsPerUnit: number
 
   tracksByObject: StrictRecord<
-    string,
+    ObjectAddressKey,
     {
       trackIdByPropPath: StrictRecord<string, SequenceTrackId>
       trackData: StrictRecord<SequenceTrackId, TrackData>
@@ -39,7 +46,10 @@ export type TrackData = BasicKeyframedTrack
 
 export type Keyframe = {
   id: KeyframeId
-  value: unknown
+  /** The `value` is the raw value type such as `Rgba` or `number`. See {@link SerializableValue} */
+  // Future: is there another layer that we may need to be able to store older values on the
+  // case of a prop config change? As keyframes can technically have their propConfig changed.
+  value: SerializableValue
   position: number
   handles: [leftX: number, leftY: number, rightX: number, rightY: number]
   connectedRight: boolean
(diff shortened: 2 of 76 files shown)


packages/dataverse/src/pointer.ts
```diff
@@ -33,9 +33,52 @@ const cachedSubPathPointersWeakMap = new WeakMap<
...
+      obj: T,
+    ): any extends T ? never[] : Extract<keyof T, string>[]
+    /** Nominal: Extension to the Object prototype definition to properly manage {@link Nominal} keyed records */
+    entries<T extends Record<Nominal<string>, any>>(
+      obj: T,
+    ): any extends T
+      ? [never, never][]
+      : Array<{[P in keyof T]: [P, T[P]]}[Extract<keyof T, string>]>
+  }
+}
```
colelawrence, line 80: Nominals and pointers ended up not really being reliant on each other's types, but they do have to work together.
I would be happy to pull the Nominals out into their own file, and perhaps move that + the Nominal specific typeTests out of dataverse.

theatre/shared/src/utils/types.ts
```diff
@@ -57,15 +64,13 @@ export type DeepPartialOfSerializableValue<T extends SerializableValue> =
...
 /**
- * This is supposed to create an "opaque" or "nominal" type, but since typescript
- * doesn't allow generic index signatures, we're leaving it be.
+ * This is equivalent to `Partial<Record<Key, V>>` being used to describe a sort of Map
+ * where the keys might not have values.
  *
- * TODO fix this once https://github.com/microsoft/TypeScript/pull/26797 lands (likely typescript 4.4)
+ * We do not use `Map`s, because they add comlpexity with converting to `JSON.stringify` + pointer types
  */
-export type Nominal<T, N extends string> = T
```
colelawrence, line 68: I'm tempted to just move Nominal back to here. WDYT, @AriaMinaei ?

packages/dataverse/src/pointer.ts
colelawrence: I moved it and rebase fixed the commits.

theatre/studio/src/uiComponents/simpleContextMenu/useContextMenu.tsx
```diff
@@ -1,31 +0,0 @@
...
-  ) : (
-    <RightClickMenu
-      items={props.items}
-      rightClickPoint={status.event}
-      onRequestClose={close}
-    />
-  )
-
-  return [node, close, status.isOpen]
-}
```
AriaMinaei, line 31: How does this differ from `usePopover()` used elsewhere?

theatre/studio/src/panels/DetailPanel/propEditors/utils/SingleRowPropEditor.tsx
```diff
@@ -109,11 +109,21 @@ const InputContainer = styled.div`
...
-}> = ({propConfig, pointerToProp, stuff, children}) => {
+}
+
+export function SingleRowPropEditor<T>({
+  propConfig,
+  pointerToProp,
+  stuff,
+  children,
+}: React.PropsWithChildren<ISingleRowPropEditorProps<T>>): React.ReactElement<
+  any,
```
AriaMinaei, line 124: `$IntentionalAny`?

packages/dataverse/src/pointer.ts
```diff
@@ -56,18 +61,19 @@ export type PointerType<O> = {
...
-    ? UnindexablePointer
-    : O extends (infer T)[]
-    ? Pointer<T>[]
-    : O extends {}
-    ? {
-        [K in keyof O]-?: Pointer<O[K]>
-      } /*&
-        {[K in string | number]: Pointer<K extends keyof O ? O[K] : undefined>}*/
-    : UnindexablePointer)
+  PointerInner<Exclude<O, undefined>, undefined extends O ? undefined : never>
```
AriaMinaei, line 64: Not sure how this works, could you explain in a comment?

theatre/core/src/projects/store/types/SheetState_Historic.ts
```diff
@@ -10,14 +18,13 @@ export interface SheetState_Historic {
...
+    byObject: StrictRecord<ObjectAddressKey, SerializableMap>
   }
-  sequence?: Sequence
+  sequence?: HistoricPositionalSequence
 }
 
-type Sequence = PositionalSequence
-
-type PositionalSequence = {
+// Question: What is this? The timeline position of a sequence?
```
AriaMinaei, line 26: Remove the comment?

theatre/core/src/propTypes/internals.ts
```diff
@@ -47,12 +48,31 @@ export type ShorthandPropToLonghandProp<P extends IShorthandProp> =
...
+    [K in keyof P]: P[K]['valueType']
+  }
+
+export type PropsValue<P> = P extends UnknownValidCompoundProps
+  ? LonghandCompoundPropsToInitialValue<P>
+  : P extends UnknownShorthandCompoundProps
+  ? LonghandCompoundPropsToInitialValue<
+      ShorthandCompoundPropsToLonghandCompoundProps<P>
+    >
+  : never
```
AriaMinaei, line 72: These can just use `ShorthandCompoundPropsToLonghandCompoundProps<P>['valueType']` and be removed

theatre/core/src/sheetObjects/SheetObject.ts
```diff
@@ -25,14 +26,28 @@ import TheatreSheetObject from './TheatreSheetObject'
...
+/**
+ * Internally, the sheet's actual configured value is not a specific type, since we
+ * can change the prop config at will, as such this is an alias of {@link SerializableMap}.
+ *
+ * TODO: Incorporate this knowledge into SheetObject & TemplateSheetObject
+ */
+type SheetObjectPropsValue = SerializableMap
+
+/**
+ * An object on a sheet consisting of zero or more properties which can be overridden and sequenced.
```
AriaMinaei, line 38: Nit `overridde`n statically or `overridde`n by being sequenced.

theatre/core/src/sheetObjects/SheetObject.ts
```diff
@@ -25,14 +26,28 @@ import TheatreSheetObject from './TheatreSheetObject'
 import type {Interpolator, PropTypeConfig} from '@theatre/core/propTypes'
 import {getPropConfigByPath} from '@theatre/shared/propTypes/utils'
 
+/**
+ * Internally, the sheet's actual configured value is not a specific type, since we
+ * can change the prop config at will, as such this is an alias of {@link SerializableMap}.
+ *
+ * TODO: Incorporate this knowledge into SheetObject & TemplateSheetObject
+ */
+type SheetObjectPropsValue = SerializableMap
```
AriaMinaei, line 35: Nice!

theatre/core/src/sheetObjects/TheatreSheetObject.ts
```diff
@@ -32,8 +34,11 @@ export interface ISheetObject<Props extends IShorthandCompoundProps = {}> {
    * const obj = sheet.object("obj", {x: 0})
    * console.log(obj.value.x) // prints 0 or the current numeric value
    * ```
+   *
+   * Question: What should we do with our type, here, if the user actually changes the
```
AriaMinaei, line 38: Right now the user can't, but we'll probably enable that functionality later (`object.overrideConfig()`). We should just educate the user that they can't rely on static types to know the type of `object.value`.

AriaMinaei (changes requested):
Left some comments. Other than those, good to merge! (also see previous review which requested `useContextMenu()` be restored)

theatre/studio/src/uiComponents/simpleContextMenu/useContextMenu.tsx
colelawrence: This was a really good thing to point out. This should have used `usePopover`, I just didn't see it. I will add JSDoc cross referencing between the two related tools.

packages/dataverse/src/pointer.ts
colelawrence: Hmm... I'll give it a whirl. First attempt to clarify: Essentially, the issue occurs is related to something like `Pointer<X | undefined>["..."]` resulting in an `UnindexablePointer`. This is because the first generic type is used to determine the overall kind of pointer (or list of pointers), but when that pointer is `| undefined` it gets off path. So, `Exclude<O, undefined>` removes any `undefined` variants in the union, and `undefined extends O ? undefined : never` results in `undefined` if `O` might be optional and `never` otherwise. So, now we can confidently use a non-undefined `O` in the ternary logic, while also ensuring that we keep track of whether it's optional via the `Optional` added param.

This is the part of the definition that matters (allowing us to remove optionality of each mapped key) without losing the information about the inner type being potentially undefined.
```ts
{
  [K in keyof O]-?: Pointer<O[K] | Optional>
}
```

theatre/core/src/projects/store/types/SheetState_Historic.ts
colelawrence: Is the comment correct?

theatre/core/src/propTypes/internals.ts
colelawrence: Note: `PropsValue<P>` is a bit more flexible than `ShorthandCompoundPropsToLonghandCompoundProps<P>`, because it accepts a dict to prop configs OR a dict to shorthand props.

theatre/core/src/sheetObjects/TheatreSheetObject.ts
colelawrence: Got it. Adding comment to this effect

packages/dataverse/src/pointer.ts
colelawrence: Applied a couple explanatory comments in the code

AriaMinaei (approved):
This is good to merge. Just the build error (caused by the rebase) is left. I'll add a commit to fix that, and merge this PR as is.

PR #149, opened 2022-05-09
Configurable ITheatreLogger

colelawrence:
A logger with three separate audiences:

 * `internal`: Logs for developers maintaining Theatre.js
 * `dev`: Logs for developers using Theatre.js
 * `public`: Logs for everyone

This logger supports:
 * multiple logging levels (error, warn, debug, trace),
 * multiple audience levels (internal, dev, public),
 * multiple categories (general, todo, troubleshooting)
 * named and keyed loggers (e.g. `rootLogger.named("Project", project.id)`)
 * console styling with deterministic coloring
 * console devtool maintains accurate sourcemap link to logging origin (e.g. `coreExports.ts:71` as opposed to `logger.ts:45` or whatever)
 * swappable logger
 * customizable filtering
 * Accepts lazy `args`: `args: () => object` via `logger.lazy.<level>("message", () => <expensive computation>)` (e.g. `logger.lazy.debugDev("Loaded project state", () => ({ save: bigProject.exportToSaveable() }))`)

### DIFF
--- a/packages/playground/src/shared/dom/index.tsx
+++ b/packages/playground/src/shared/dom/index.tsx
@@ -3,6 +3,7 @@ import ReactDOM from 'react-dom'
 import studio from '@theatre/studio'
 import {getProject} from '@theatre/core'
 import {Scene} from './Scene'
+import {TheatreLoggerLevel} from '@theatre/shared/logger'
 /**
  * This is a basic example of using Theatre for manipulating the DOM.
  *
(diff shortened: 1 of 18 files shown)


colelawrence:
I've fixed the remaining extraneous log in a fixup. This should be good to go. If there are any concerns with how this affects bundle size of core, we can definitely apply some optimizations for how it gets bundled to remove internal logging statements via esbuild plugins.

AriaMinaei (approved):
I’m 9/10 on the utility of the logger, 5/10 on the implementation details, mainly because I haven’t used it enough to form an opinion. I think it’s more productive not to wait for me to form that opinion. We can merge it in as an experiment, gain more experience with it. Since it’s an experiment, we can improve/remove later or most likely keep as is.

PR #151, opened 2022-05-09
Fix curveEditorPopover overwriting existing curve

### LINKED ISSUE #148: `SingleTweenEditor` overrides the existing tween as soon as the popover opens
Screen recording:

https://user-images.githubusercontent.com/593118/167151223-fae38a7b-b521-4fa2-9b6a-fdd91263774a.mp4

vezwork:
Fixes #148

### DIFF
--- a/theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/CurveEditorPopover.tsx
+++ b/theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/CurveEditorPopover.tsx
@@ -154,14 +154,16 @@ const CurveEditorPopover: React.FC<IProps> = (props) => {
     inputRef.current?.focus()
   }, [inputRef.current])
 
-  const [inputValue, setInputValue] = useState<string>('')
+  const [inputValue, setInputValue] = useState<string>(
+    cssCubicBezierArgsFromHandles(easing),
+  )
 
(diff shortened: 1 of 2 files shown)


theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/BasicKeyframedTrack/KeyframeEditor/CurveEditorPopover/CurveEditorPopover.tsx
```diff
@@ -197,11 +198,10 @@ const CurveEditorPopover: React.FC<IProps> = (props) => {
 
   // When `preview` or `edit` change, use the `tempTransaction` to change the
   // curve in Theate's data.
-  useMemo(
-    () =>
-      setTempValue(tempTransaction, props, cur, next, preview ?? edit ?? ''),
-    [preview, edit],
-  )
+  useMemo(() => {
```
AriaMinaei, line 201: Catching this just now. This should be a `useEffect()` rather than `useMemo()` as it's not returning any value and is running a side-effect.

I'd use `useLayoutEffect()` to ensure no delay between first render and the effect.

AriaMinaei (changes requested):

vezwork:
@AriaMinaei I caught and fixed a few additional things:
- I realized `TextInputMode` was not necessary, it is removed. Code is simpler now.
- My logic for reacting to changes to the input was incorrect (the callback looked at the previously bound value of `inputValue` instead of `e.target.value`.
- When selecting an item in the grid it getting unselected. 
- One or two other very minor things.

PR #152, opened 2022-05-10
Refactor `useDrag()` and implement middle button dragging

AriaMinaei:
This PR simplifies `useDrag()`'s API, and fixes various scrolling issues on windows, mainly the bug that appears when scrolling small amounts via mouse/trackpad would cause SequenceEditor's horizontal scroll to drift without actually updating the clipped space range.


vezwork (approved):
This change makes sense to me 👍

PR #153, opened 2022-05-10
Support multiple sheet instances

AndrewPrifer:
This PR adds support for multiple sheet instances of the same sheet, enabling creating editable components backed by the same theatre sheet.

Aside from that, the PR also slightly modifies the SheetProvider API, changing the getSheet prop to simply sheet, allowing passing in the sheet itself directly, while also ensuring that the sheet context always provides a value from the very first render.


packages/r3f/src/utils.ts
```diff
@@ -1,3 +1,7 @@
 import {useEditorStore} from './store'
+import type {ISheet} from '@theatre/core'
 
 export const refreshSnapshot = useEditorStore.getState().createSnapshot
+
+export const makeObjectKey = (sheet: ISheet, name: string) =>
+  `${sheet.address.sheetId}:${sheet.address.sheetInstanceId}:${name}`
```
AriaMinaei, line 7: I was confused about the word "objectKey" as that's also the word we use in `@theatre/core` for `sheet.object(key, ...)`. Maybe rename it to `makeStoreKey()`?

packages/r3f/src/components/editable.tsx
```diff
@@ -47,7 +48,9 @@ const createEditable = <Keys extends keyof JSX.IntrinsicElements>(
 
         const objectRef = useRef<JSX.IntrinsicElements[U]>()
 
-        const sheet = useCurrentSheet()
+        const sheet = useCurrentSheet()!
+
+        const objectKey = makeObjectKey(sheet, uniqueName)
```
AriaMinaei, line 53: Smae as the other one. Perhaps `storeKey` is less confusing.

AriaMinaei (changes requested):
Requested a minor change. Other than that, good to merge!

packages/r3f/src/components/editable.tsx
```diff
@@ -140,7 +140,7 @@ const createEditable = <Keys extends keyof JSX.IntrinsicElements>(
             visible={visible !== 'editor' && visible}
             userData={{
               __editable: true,
-              __editableName: objectKey,
+              __editableName: storeKey,
```
AriaMinaei, line 143: Probably should rename `__editableName` too, right?

AndrewPrifer: True

PR #143, opened 2022-05-04
Publish the development versions of the packages as tarballs

fulopkovacs:
# Description

> :warning: We decided to test publishing the dev packages to npm instead of an object storage. This PR is likely to be closed without being merged!

This PR adds a script that generates tarballs from the `@theatre` packages and publishes them in the CI. [See this  (private) Notion doc](https://www.notion.so/theatrejs/POC-Publishing-development-packages-2e282feb3e85408aac9738eeb0a53e72) for the behind-the-scenes information.

# Versioning of the dev packages
A CI job builds the packages and assigns them a version number based on the version (without tags like `-dev.3`) specified in the root `package.json` (or `packages/r3f/package.json` in the case of `r3f` since that has an independent release schedule) and the abbreviated hash of the latest commit. Here's an example:
| prop | value |
| --- | --- |
| package name | `@theatre/core` |
| root version | `0.4.8-dev.3` |
| latest commit hash | `c8884093` |
| dev package version | `0.4.8-insiders.c8884093` |
| url (the domain will change to `packages.theatrejs.com`, or something similiar in the future) | `https://theatrejs-packages.nyc3.cdn.digitaloceanspaces.com/@theatre/core-c8884093.tgz` |
| install command | `yarn add https://theatrejs-packages.nyc3.cdn.digitaloceanspaces.com/@theatre/core-c8884093.tgz` 

# Testing this PR

1. Clone this repo: https://github.com/fulopkovacs/theatre-r3f-create-react-app
```
git clone git@github.com:fulopkovacs/theatre-r3f-create-react-app.git
```
2. Run the following commands in the repo (the second command takes quite a while):
```
yarn && yarn start
```
3. Visit `http://localhost:3000` in the browser
_Optional_:  You can verify that you're using the dev packages by checking the version of the `@theatre` packages with `yarn why <package-name>`, like:
```
$ yarn why @theatre/core
yarn why v1.22.10
[1/4] Why do we have the module "@theatre/core"...?
[2/4] Initialising dependency graph...
[3/4] Finding dependency...
[4/4] Calculating file sizes...
=> Found "@theatre/core@0.4.8-insiders.c8884093"
info Has been hoisted to "@theatre/core"
info This module exists because it's specified in "dependencies".
info Disk size without dependencies: "656KB"
info Disk size with unique dependencies: "1.13MB"
info Disk size with transitive dependencies: "3.73MB"
info Number of shared dependencies: 2
Done in 0.42s.
```

[...]

PR #155, opened 2022-05-11
Publish development packages to npm

fulopkovacs:
# Description

Build the `@theatre` packages based on the latest commit of the PR and publish them to npm (as `insider` packages).

# Versioning of the dev packages
A CI job builds the packages and assigns them a version number based on the version (without tags like `-dev.3`) specified in the root `package.json` (or `packages/r3f/package.json` in the case of `r3f` since that has an independent release schedule) and the abbreviated hash of the latest commit. Here's an example:
| prop | value |
| --- | --- |
| package name | `@theatre/core` |
| root version | `0.4.8-dev.3` |
| latest commit hash | `a369b71` |
| dev package version | `0.4.8-insiders.a369b71` |
| install command | `yarn add @theatre/core@0.4.8-insiders.a369b71` | 

# Testing this PR

1. Clone this repo: https://github.com/fulopkovacs/theatre-r3f-create-react-app
```
git clone git@github.com:fulopkovacs/theatre-r3f-create-react-app.git
```
2. Run the following commands in the repo (the second command takes quite a while):
```
yarn && yarn start
```
3. Visit `http://localhost:3000` in the browser
_Optional_:  You can verify that you're using the dev packages by checking the version of the `@theatre` packages with `yarn why <package-name>`, like:
```
$ yarn why @theatre/core
yarn why v1.22.10
[1/4] Why do we have the module "@theatre/core"...?
[2/4] Initialising dependency graph...
[3/4] Finding dependency...
[4/4] Calculating file sizes...
=> Found "@theatre/core@0.4.8-insiders.a369b71"
info Has been hoisted to "@theatre/core"
info This module exists because it's specified in "dependencies".
info Disk size without dependencies: "656KB"
info Disk size with unique dependencies: "1.13MB"
info Disk size with transitive dependencies: "3.73MB"
info Number of shared dependencies: 2
Done in 0.42s.
```

# CodeSandbox Example
Here's a CodeSandbox that uses the `insider` versions of the `@theatre packages`:
https://codesandbox.io/s/theater-js-r3f-using-insider-packages-2nvre0

# TODO after this PR is merged
Close https://github.com/theatre-js/theatre/pull/143.


fulopkovacs:
@AriaMinaei  Here are a few notes about this PR:

- First, IMO this should be ready for the merge.
- Ideally we would have GH actions that edits the same comment under the PR in the following way:
    1. One action writes an "In Progress"-message when the CI starts
    2. Another action writes a "Packages have been published, here are the versions: ..."-message if the packages were published successfully
    3. And one that writes a "Failed to publish packages"-message if we couldn't publish them.
   - I've been working on this feature since this morning, but not sure how much more time would it take. It's surprisingly hard to pass information around about the published packages between jobs. At this point I'd argue that we could drop it, and come back to this problem in the future, when we can afford to spend a day on researching solutions and another one to implementing the best one.
- I separated the tests from publishing the packages into two different jobs. This added about 1 min to the whole workflow, because that's how much the `linking` process takes when we install the dependencies again for the second job (I couldn't find a way to share runners between the jobs, [logs](https://github.com/theatre-js/theatre/runs/6406822272?check_suite_focus=true#step:6:51)):
image

fulopkovacs:
@AriaMinaei Apparently, all the workflows that are triggered from a forked repo might fail, because they won't be able to publish the packages:
> With the exception of GITHUB_TOKEN, secrets are not passed to the runner when a workflow is triggered from a forked repository. ([source](https://docs.github.com/en/actions/security-guides/encrypted-secrets#using-encrypted-secrets-in-a-workflow))

GitHub also doesn't seem to support allowing a workflow step to fail: https://github.com/actions/toolkit/issues/399. We can continue the workflow after an error, but will never get a green check ([people came up with  workarounds, but those have limitations](https://github.com/actions/toolkit/issues/399#issuecomment-1103775641))

PR #135, opened 2022-04-27
wip: Inline keyframe editing

colelawrence:
Left click a keyframe to "open" it for editing.

PR #157, opened 2022-05-13
Compound prop context menu

AriaMinaei:
Will have to rebase on #135


colelawrence:
Agreed to hold until #135, once #135 is merged, we could do a live-share pair & review to re-implement these changes on top. My gut tells me we should just do the rebase by hand (checking out main and writing these code changes into the codebase again with this PR as reference as opposed to `git rebase`).

PR #166, opened 2022-05-16
Expose editable factory API

AndrewPrifer:
This PR makes creating a custom editable possible from the userland.

With the help of this API, you can support any object with any property you want, even ones that are not part of the scene graph! In fact, this is the API we use for the default internal config too!

API additions:
- `createEditable<Keys extends keyof JSX.IntrinsicElements>(config: EditableFactoryConfig<Keys>)`: creates a custom `editable` utility using the provided config
- `defaultEditableConfig`: the default config, you can extend this to create your own config, or you can create one from scratch
-  `createColorPropConfig()`, `createNumberPropConfig()`, `createVectorPropConfig()`: utilities for defining props in the config


AriaMinaei:
Closing after our internal meeting. Gonna give this idea a bit more time so we can figure out the edge cases, like "how to handle deeply nested compounds."

AriaMinaei:
Closing after our internal meeting. Gonna give this idea a bit more time so we can figure out the edge cases, like "how to handle deeply nested compounds."

PR #167, opened 2022-05-16
Feature/2022 05/collapsing in dopesheet

colelawrence:
Add a feature for collapsing rows and compound props in the "Sequence Editor".

https://user-images.githubusercontent.com/2925395/168878519-9234ff42-3005-4406-8a90-b83083eb743c.mp4


vezwork:
Note: just noticed that when shift+dragging to select, the keyframe bounding box is offset to the left. This is a problem(?) on `main`.

theatre/studio/src/panels/SequenceEditorPanel/layout/tree.ts
```diff
@@ -105,29 +113,35 @@ export const calculateSequenceEditorTree = (
...
-    addProps(
-      sheetObject,
-      trackSetups,
-      [],
-      sheetObject.template.config,
-      row.children,
-      level + 1,
-    )
-
+    topSoFar += HEIGHT_OF_ANY_TITLE
```
colelawrence, line 134: Suggestion: add comments for what each addition is, in this case it's for the height of the sheet object row header, I think, right?

colelawrence (commented):
Small feedback.

PR #168, opened 2022-05-16
Feature/2022 05/sheet object namespacing in outline

colelawrence:


vezwork:
note: I have removed the draft changes that add namespacing to the sequence editor. I still have the changes on a local branch if we need them for reference later.

vezwork (approved):
@colelawrence ready for merge IMO.

PR #178, opened 2022-05-26
Single tween editor for aggregate rows

vezwork:
- added background curves to multiple value curve editor

https://user-images.githubusercontent.com/11082236/170716407-b0ec2a4c-2a0b-45fd-841c-c98108a4c2a9.mp4

image

## notes

- behaviour: selected connections are always edited when you click on a aggregate connection if they share a track with the aggregate. This may be unexpected because the selection can be in a different part of the timeline than the aggregate. This behaviour occurs because the aggregate connector and basic connector share the same selection logic, but the aggregate connector is passed, as a prop, more of the selection than the basic connector.

Co-authored-by: Cole Lawrence <cole@colelawrence.com>

### DIFF
--- a/theatre/studio/src/store/stateEditors.ts
+++ b/theatre/studio/src/store/stateEditors.ts
@@ -777,8 +777,6 @@ namespace stateEditors {
           /**
            * Sets the easing between keyframes
            *
-           *
-           *
            * X = in keyframeIds
            * * = not in keyframeIds
            * + = modified handle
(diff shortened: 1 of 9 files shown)


theatre/studio/src/store/stateEditors.ts
```diff
@@ -828,6 +838,31 @@ namespace stateEditors {
...
+            if (!track) return
+            track.keyframes = track.keyframes.map((kf) => {
+              if (kf.id === p.keyframeId) {
+                return {
+                  ...kf,
+                  handles: [
+                    p.end?.[0] ?? kf.handles[0],
+                    p.end?.[1] ?? kf.handles[1],
+                    p.start?.[0] ?? kf.handles[2],
+                    p.start?.[1] ?? kf.handles[3],
```
colelawrence, line 859: Add docs like: `Use given value or fallback to original value, allowing the caller to customize exactly which side of the curve they are editing.`

colelawrence (changes requested):

AriaMinaei (approved):

PR #181, opened 2022-05-30
Dev/2022 05/studio frp (experimental mvvm package)

colelawrence:
See #188 as well.

This PR includes some interesting tools borrowed from `rxjs` for push-based data-flow reactivity and disposable management.

I liked the minor improvements I could make to the original naming/setup in rxjs and I appreciated removing some of their surface area I consider cruft like error/complete in the context of MVVM implementation. Even so, the vast surface area of tooling and documentation of rxjs, makes it still hard to imagine trying to compete with how well tested and optimized they already are. There are ways to optimize rxjs further, but it's ultimately 90% of the way there, and it is a herculian task to try to catch up with all the utilities and testing they already provide.

So, if we wanted to further investigate a push model of updating UI, then I would seriously consider just using rxjs directly as an alternative if the needs are there.


colelawrence:
Updated comment. Not of great relevance currently

PR #179, opened 2022-05-27
Interactive aggregate keyframes

AndrewPrifer:
This PR extends the current, visual-only implementation of aggregate keyframes with interactivity, i.e. aggregate keyframes will properly handle dragging and associated selection logic.

PR #183, opened 2022-05-30
Feature/2022 05/aggregate refactor and keyframe drag

colelawrence:
Supercedes #179 & includes minor hotfixes and refatoring for AggregateKeyframe organization


theatre/studio/src/uiComponents/Popover/usePopover.tsx
```diff
@@ -88,6 +93,9 @@ export default function usePopover(
               threshold: opts.pointerDistanceThreshold ?? 100,
               callback: () => {
                 if (lock.childHasFocusRef.current) return
+                // this is a bit weird, because when you stop capturing, then the popover can close on you...
+                // TODO: Better fixes?
```
AndrewPrifer, line 97: Only way to fix this corner case imo would be to use pointer lock instead of pointer capture. Locking would keep the pointer in place, not just eliminating this corner case, but entirely eliminating the need for this logic. But locking can be buggy (see number input) so probably not for now.

AndrewPrifer (approved):

AndrewPrifer (approved):

AndrewPrifer (approved):
LGTM

PR #189, opened 2022-06-01
Fix aggregate keyframe dragging stopping in an edge case when the key for the drag element changes

AndrewPrifer:
We fixed this by using the container, which is stable as the drag element.


colelawrence (approved):
LGTM. Great work!

PR #195, opened 2022-06-04
Perfectly precise selections

AndrewPrifer:
- [x] Fix selection box misalignment
- [x] Improve keyframe selection hit-zone dimensions
- [ ] Document code

### Fix selection box misalignment
This fix involves correctly converting being normal space and left-padded space (normal space + left padding). The distinction between padded and normal on the space-level is important so that we don't have to go and ad-hoc patch every single calculation with the padding amount. Removing the padding in the drag handler and adding it back in the selection box might seem round-about, but this way we can perform every other calculation without conversions.

### Improve keyframe selection hit-zone dimensions
Currently, the hit-zones of the keyframes are represented by a vertical line spanning the entire track. This means, that you have to be needlessly precise and make sure you position the selection box only in the tracks in which you want to select keyframes on the vertical axis, while on the horizontal axis, the hit-zone is point-like, requiring the center of a keyfame to be covered by the selection in order to be included. The implementation in this PR aligns the hit-zone more closely with the visual appearance of the keyframe, both making selections more intuitive and requiring less precision from the user.

### DIFF
--- a/theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/DopeSheetSelectionView.tsx
+++ b/theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/DopeSheetSelectionView.tsx
@@ -25,6 +25,8 @@ import {collectAggregateKeyframesInPrism} from './collectAggregateKeyframes'
 import type {ILogger, IUtilLogger} from '@theatre/shared/logger'
 import {useLogger} from '@theatre/studio/uiComponents/useLogger'
 
+const HITBOX_SIZE_PX = 5
+
 const Container = styled.div<{isShiftDown: boolean}>`
   cursor: ${(props) => (props.isShiftDown ? 'cell' : 'default')};
 `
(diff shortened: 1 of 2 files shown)


theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/DopeSheetSelectionView.tsx
```diff
@@ -50,13 +52,22 @@ const DopeSheetSelectionView: React.FC<{
...
+  h: [from: number, to: number]
+  /**
+   * The vertical bounds of the selection as a tuple of "from" and "to" coordinates, "from" representing the start of the drag.
+   */
+  v: [from: number, to: number]
 }
 
 function useCaptureSelection(
-  layoutP: Pointer<SequenceEditorPanelLayout>,
+  layout: SequenceEditorPanelLayout,
```
AriaMinaei, line 70: using `layout` instead of `layoutP` is going to make this hook re-evaluate its `useMemo` very time `layout` changes when the selecting drag gesture is not running.

AriaMinaei (changes requested):

PR #197, opened 2022-06-07
Jump between the sequence editor panel and the details panel

AriaMinaei:


colelawrence:
@AriaMinaei, I've rebased on the latest from `dev/introduce-initial-derive-helpers` to this branch, so there shouldn't be any issues if you `git pull -r` down and make the necessary tweaks.

PR #199, opened 2022-06-07
Support the three/four/six-value syntax variants of <hex-color>

fulopkovacs:
This PR adds support for the following syntax variants of <hex-color> (see the [docs](https://developer.mozilla.org/en-US/docs/Web/CSS/hex-color#syntax)):

```
#RGB        // The three-value syntax
#RGBA       // The four-value syntax
#RRGGBB     // The six-value syntax
#RRGGBBAA   // The eight-value syntax (we already support this)
```

The PR also fixes some code style inconsistencies in AggregateKeyframeTrack.tsx.


fulopkovacs:
It turned out that we have quite a few redundant lines of code around the color prop. I decided to address this problem in a different PR (to keep things clean): https://github.com/theatre-js/theatre/pull/201

theatre/studio/src/propEditors/simpleEditors/RgbaPropEditor.tsx
```diff
@@ -1,5 +1,6 @@
 import type {PropTypeConfig_Rgba} from '@theatre/core/propTypes'
-import type {Rgba} from '@theatre/shared/utils/color'
+import type {Rgba} from '@theatre/shared/utils/color';
+import { validHexRegExp} from '@theatre/shared/utils/color'
```
AriaMinaei, line 3: I'm surprised the commit hook isn't picking up these style inconsistencies. Or did you run `git commit --no-verify`? ;)

fulopkovacs: Nope, I simply ran `git commit`. Interesting indeed. I'll fix this one (and will check if there's more).

PR #201, opened 2022-06-07
Remove unused lines of code around the color prop

fulopkovacs:
There are a lot of lines of code around the color prop that are not used at all. Some comments in the code base suggest that it's in need for a refactor, but instead of changing everything at once I think we can start by just removing the unused lines. We can refactor the code later. :relaxed:


AndrewPrifer:
I think we can go ahead with merging this, although let's make a mental note of this PR since we might have to add back some of the utilities as our toolset expands.

PR #202, opened 2022-06-07
Add initial tools for managing derivations and React compatibility

colelawrence:
Initial tools for using dataverse derivations in more places to enable fewer re-renders.

Foundation for #197

### DIFF
--- a/theatre/studio/src/utils/derive-utils.tsx
+++ b/theatre/studio/src/utils/derive-utils.tsx
@@ -52,7 +52,8 @@ function DeriveElement(props: {der: IDerivation<React.ReactNode>}) {
   return <>{node}</>
 }
 
-class TSErrors<M> extends Error {}
+/** This is only used for type checking to make sure the APIs are used properly */
+interface TSErrors<M> extends Error {}
 
 type ReactDeriver<Props extends {}> = (
   props: {
--- a/theatre/studio/src/utils/derive-utils.tsx
+++ b/theatre/studio/src/utils/derive-utils.tsx
@@ -1,0 +1,109 @@
+import {isDerivation, prism, val} from '@theatre/dataverse'
+import type {IDerivation, Pointer} from '@theatre/dataverse'
+import {useDerivation} from '@theatre/react'
+import type {$IntentionalAny} from '@theatre/shared/utils/types'
+import React, {useMemo, useRef} from 'react'
+import {invariant} from './invariant'
+
+type DeriveAll<T> = IDerivation<
+  {
+    [P in keyof T]: T[P] extends $<infer R> ? R : never
+  }
+>
+
+export type $<T> = IDerivation<T> | Pointer<T>
+
+function deriveAllD<T extends [...$<any>[]]>(obj: T): DeriveAll<T>
+function deriveAllD<T extends Record<string, $<any>>>(obj: T): DeriveAll<T>
+function deriveAllD<T extends Record<string, $<any>> | $<any>[]>(
+  obj: T,
+): DeriveAll<T> {
+  return prism(() => {
+    if (Array.isArray(obj)) {
+      const values = new Array(obj.length)
+      for (let i = 0; i < obj.length; i++) {
+        values[i] = obj[i].getValue()
+      }
+      return values
+    } else {
+      const values: $IntentionalAny = {}
+      for (const k in obj) {
+        values[k] = val((obj as Record<string, $<any>>)[k])
+      }
+      return values
+    }
+  }) as $IntentionalAny
+}
+
+export function useReactPrism(
+  fn: () => React.ReactNode,
+  deps?: any[],
+): React.ReactElement {
+  const derivation = useMemo(() => prism(fn), deps ?? [])
+  return <DeriveElement der={derivation} />
+}
+
+export function reactPrism(fn: () => React.ReactNode): React.ReactElement {
+  return <DeriveElement der={prism(fn)} />
+}
+
+function DeriveElement(props: {der: IDerivation<React.ReactNode>}) {
+  const node = useDerivation(props.der)
+  return <>{node}</>
+}
+
+class TSErrors<M> extends Error {}
+
+type ReactDeriver<Props extends {}> = (
+  props: {
+    [P in keyof Props]: Props[P] extends IDerivation<infer _>
+      ? TSErrors<"Can't both use Derivation properties while wrapping with deriver">
+      : Props[P] | IDerivation<Props[P]>
+  },
+) => React.ReactElement | null
+
+/**
+ * Wrap up the component to enable it to take derivable properties.
+ * Invoked similarly to `React.memo`.
+ */
+export function deriver<Props extends {}>(
+  Component: React.ComponentType<Props>,
+): ReactDeriver<Props> {
+  return React.forwardRef(function deriverRender(
+    props: Record<string, $IntentionalAny>,
+    ref,
+  ) {
+    let observableArr = []
+    const observables: Record<string, IDerivation<$IntentionalAny>> = {}
+    const normalProps: Record<string, $IntentionalAny> = {
+      ref,
+    }
+    for (const key in props) {
+      const value = props[key]
+      if (isDerivation(value)) {
+        observableArr.push(value)
+        observables[key] = value
+      } else {
+        normalProps[key] = value
+      }
+    }
+
+    const initialCount = useRef(observableArr.length)
+    invariant(
+      initialCount.current === observableArr.length,
+      `expect same number of observable props on every invocation of deriver wrapped component.`,
+      {initial: initialCount.current, count: observableArr.length},
+    )
+
+    const allD = useMemo(() => deriveAllD(observables), observableArr)
+    const observedPropState = useDerivation(allD)
+
+    return (
+      observedPropState &&
+      React.createElement(Component, {
+        ...normalProps,
+        ...observedPropState,
+      } as Props)
+    )
+  })
+}
--- a/theatre/studio/src/utils/derive-utils.tsx
+++ b/theatre/studio/src/utils/derive-utils.tsx
@@ -4,6 +4,7 @@ import {useDerivation} from '@theatre/react'
 import type {$IntentionalAny} from '@theatre/shared/utils/types'
 import React, {useMemo, useRef} from 'react'
 import {invariant} from './invariant'
+import {emptyArray} from '@theatre/shared/utils'
 
 type DeriveAll<T> = IDerivation<
   {
@@ -37,9 +38,9 @@ function deriveAllD<T extends Record<string, $<any>> | $<any>[]>(
 
 export function useReactPrism(
   fn: () => React.ReactNode,
-  deps?: any[],
+  deps: readonly any[] = emptyArray,
 ): React.ReactElement {
-  const derivation = useMemo(() => prism(fn), deps ?? [])
+  const derivation = useMemo(() => prism(fn), deps)
   return <DeriveElement der={derivation} />
 }
 
--- a/theatre/studio/src/utils/derive-utils.tsx
+++ b/theatre/studio/src/utils/derive-utils.tsx
@@ -66,6 +66,11 @@ type ReactDeriver<Props extends {}> = (
 /**
  * Wrap up the component to enable it to take derivable properties.
  * Invoked similarly to `React.memo`.
+ *
+ * @remarks
+ * This is an experimental interface for wrapping components in a version
+ * which allows you to pass in derivations for any of the properties that
+ * previously took only values.
  */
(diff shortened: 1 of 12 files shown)


theatre/studio/src/utils/derive-utils.tsx
```diff
@@ -0,0 +1,117 @@
...
+      ? TSErrors<"Can't both use Derivation properties while wrapping with reactDerived">
+      : Props[P] | IDerivation<Props[P]>
+  },
+) => React.ReactElement | null
+
+/**
+ * Wrap up the component to enable it to take derivable properties.
+ * Invoked similarly to `React.memo`.
+ */
+export function reactDerived<Props extends {}>(
```
AndrewPrifer, line 77: - Naming nitpick: technically, the component is not the one that is derived, it just observes its props. We don't have a term for observers in the dataverse lingo yet though, but changing it to reflect that would be nicer.

- I'm not a fan of namespacing in export names. As long as the name is unambiguous in the module context, it's fine, we can always alias it at import. `reactDerived` is also easily confused with `deriveReact`.

AndrewPrifer (commented):
I'm not yet super sure which APIs should be exported for now. I'd need to see more code examples of how and in what context these could be used to have a proper opinion, since these APIs look like they can be used interchangeable at a lot of places.

Let's also add proper docs so it's easy for everyone to pick these utilities up.

AndrewPrifer (approved):

vezwork:
Very cool PR

+1 to a rename of `derivedChild`. `deriveEl` seems okay (although, does it work with any element or only Styled ones? Maybe it should be called `deriveStyledEl`? Or `styledElFromDerivedProps`?)

theatre/studio/src/utils/tightJsonStringify.ts
```diff
@@ -0,0 +1,9 @@
+export function tightJsonStringify(
+  obj: any,
+  replacer?: ((this: any, key: string, value: any) => any) | undefined,
+) {
+  return JSON.stringify(obj, replacer, 2)
+    .replace(/^([\{\[])\n (\s+)/, '$1$2')
+    .replace(/(\n[ ]+[\{\[])\n\s+/g, '$1 ')
+    .replace(/\n\s*([\]\}])/g, ' $1')
+}
```
vezwork, line 9: This needs docs / examples or named regexs. Why would this be used instead of JSON.stringify? What does `tight` mean? No whitespace?

theatre/studio/src/utils/stringify.ts
```diff
@@ -0,0 +1,23 @@
...
+  }
+}
+
+export function stringifyError(err: any) {
+  return err instanceof Error
+    ? err.toString()
+    : err == null
+    ? undefined
+    : stringify(err)
+}
```
vezwork, line 23: This has a very general name. Is it intended to be THE stringify function we use? Or does it have a particular purpose? Can you make the name more specific? Why this stringify function instead of just JSON.stringify?

Please add jsdoc.

theatre/studio/src/utils/stringify.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export function stringify(input: any, indentJSON: boolean = true): string {
+  try {
+    return typeof input === 'string'
+      ? input
+      : typeof input === 'function'
+      ? input.toString()
+      : indentJSON
+      ? tightJsonStringify(input)
+      : JSON.stringify(input)
```
vezwork, line 11: I find these ternary chains hard to read because I'm not used to them and because of the lack of difference in indentation between the conditions and results. I would find this much easier to read if this was broken into nested ifs with named conditions or if else chains or something.

theatre/studio/src/utils/invariant.ts
```diff
@@ -0,0 +1,72 @@
+import {stringify} from './stringify'
+
+type accepts = string | number | object
+export function invariant(
```
vezwork, line 4: Needs docs. What is an invariant? It seems like this is some sort of an assertion? Is this for adding tests inline in our code? How is it used? I cannot exactly tell what sort of output this would produce in what cricumstances since the body of the function is quite general.

theatre/studio/src/utils/invariant.ts
```diff
@@ -0,0 +1,72 @@
...
+  throw new InvariantError(
+    'Invariant: ' +
+      stringify(typeof message === 'function' ? message() : message) +
+      (arguments.length > 1
+        ? `\nInstead found: ${
+            // check if probably already contains stringified json
+            /": "/.test(found) ? found : stringify(found)
+          }`
+        : ''),
+    found,
```
vezwork, line 36: Why is this code so similar but slightly different to the body of `invariant`? Can it be refactored?  Can you pull out some of these expressions into variables with expressive names?

theatre/studio/src/utils/invariant.ts
```diff
@@ -0,0 +1,72 @@
...
+  )
+}
+
+// remove these lines from thrown errors
+const AT_NODE_INTERNAL_RE = /^\s*at.+node:internal.+/gm
+const AT_INVARIANT_RE = /^\s*(at|[^@]+@) (?:Object\.)?invariant.+/gm
+const AT_TEST_HELPERS_RE = /^\s*(at|[^@]+@).+test\-helpers.+/gm
+// const AT_WEB_MODULES = /^\s*(at|[^@]+@).+(web_modules|\-[a-f0-9]{8}\.js).*/gm
+const AT_ASSORTED_HELPERS_RE =
+  /^\s*(at|[^@]+@).+(debounce|invariant|iif)\.[tj]s.*/gm
```
vezwork, line 53: Why remove these lines? What errors are we talking about? All errors or specifically the invariant ones?

theatre/studio/src/utils/invariant.ts
```diff
@@ -0,0 +1,72 @@
+import {stringify} from './stringify'
```
vezwork, line 1: I do not understand this file. The implementations and types are terse and I don't understand how it is intended to be used.

theatre/studio/src/utils/derive-utils.tsx
```diff
@@ -0,0 +1,109 @@
...
+export function reactPrism(fn: () => React.ReactNode): React.ReactElement {
+  return <DeriveElement der={prism(fn)} />
+}
+
+function DeriveElement(props: {der: IDerivation<React.ReactNode>}) {
+  const node = useDerivation(props.der)
+  return <>{node}</>
+}
+
+class TSErrors<M> extends Error {}
```
vezwork, line 55: Why define this type? Isn't it the same as `Error`? Needs a doc?

theatre/studio/src/utils/derive-utils.tsx
```diff
@@ -0,0 +1,109 @@
...
+      ? TSErrors<"Can't both use Derivation properties while wrapping with deriver">
+      : Props[P] | IDerivation<Props[P]>
+  },
+) => React.ReactElement | null
+
+/**
+ * Wrap up the component to enable it to take derivable properties.
+ * Invoked similarly to `React.memo`.
+ */
+export function deriver<Props extends {}>(
```
vezwork, line 69: I found this function to be nice to read. I don't quite understand why an invariant is used but otherwise it is very easy to understand.

vezwork (changes requested):

theatre/studio/src/utils/stringify.ts
colelawrence: Good questions, this, is not `JSON.stringify`, this is the "I need to create a displayable string for errors and logging of some arbitrary objects" function.

theatre/studio/src/utils/stringify.ts
colelawrence: We can add comments for helping the distinguishing between what is a boolean check versus value.

How long did you look at this? Was it confusing after a few seconds, or really hard to reason about due to not seeing this pattern much before?

e.g. is this a Nit or a Suggestion?

theatre/studio/src/utils/invariant.ts
colelawrence: This is an invariant helper [see examples](https://javascript.hotexamples.com/examples/tiny-invariant/-/default/javascript-default-function-examples.html).

It is a really useful building block which provides its own Error which can clean up it's stack trace correctly for cases like in test files. All these helpers are really valuable carry-overs from another large codebase, but there's a lot to talk about with why these are each large QoL improvements over out-of-the box tools seen in jest/built-in functions.

theatre/studio/src/utils/invariant.ts
colelawrence: We might be able to add another creator function to InvariantError, but generally, the rationale is to ensure that the stack trace is kept minimal.

theatre/studio/src/utils/invariant.ts
colelawrence: These lines do not contribute to understanding where the actual error comes from, so they obfuscate the call chain.
This has the greatest effect in jest, where you really want to get only the truly significant lines of code in your stack trace.

theatre/studio/src/utils/stringify.ts
vezwork: I will say this is a nit. I still prefer if elses becaue I think it is more accessible for people, but I realized this is actually not that bad to read.

vezwork (approved):

vezwork:
I like the changes to dataverse too

colelawrence:
Re-requested Aria for the review of dataverse changes.

AriaMinaei (changes requested):

theatre/studio/src/utils/derive-utils.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+export function useReactPrism(
+  fn: () => React.ReactNode,
+  deps?: any[],
+): React.ReactElement {
+  const derivation = useMemo(() => prism(fn), deps ?? [])
+  return <DeriveElement der={derivation} />
+}
+
+export function reactPrism(fn: () => React.ReactNode): React.ReactElement {
```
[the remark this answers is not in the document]
colelawrence: This is the non-memoized version of this. For example, this is used by "keyframe control indicators" which creates a react element which does not actually exist within a hook context.

AriaMinaei:
@colelawrence couple of unresolved comments remain. Feel free to merge after addressing them.

theatre/studio/src/utils/invariant.ts
```diff
@@ -0,0 +1,93 @@
+import {devStringify} from './devStringify'
+
+type accepts = string | number | object
+
+/**
+ *
+ */
+export function invariant(
```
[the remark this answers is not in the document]
colelawrence: That screws up the type system. We can instead just move that check to within the function, here. WDYT?

theatre/studio/src/utils/devStringify.ts
```diff
@@ -0,0 +1,15 @@
+import {tightJsonStringify} from './tightJsonStringify'
+
+export function devStringify(input: any, indentJSON: boolean = true): string {
```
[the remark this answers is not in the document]
colelawrence: Applied by @vezwork

theatre/studio/src/utils/derive-utils.tsx
colelawrence: Applied comments and made this an `interface` so it's not included

theatre/studio/src/utils/derive-utils.tsx
```diff
@@ -0,0 +1,109 @@
...
+      return values
+    }
+  }) as $IntentionalAny
+}
+
+export function useReactPrism(
+  fn: () => React.ReactNode,
+  deps?: any[],
+): React.ReactElement {
+  const derivation = useMemo(() => prism(fn), deps ?? [])
```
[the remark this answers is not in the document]
colelawrence: Applied

theatre/studio/src/utils/invariant.ts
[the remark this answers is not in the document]
colelawrence: I mean to say that invariant `asserts` the value in TypeScript's type space, so calls after this can actually expect that invariant condition is guaranteed. This benefit is removed if we have to wrap it in a separate node env check.

Additionally, adding env guards can make the code less readable.

I'd like to see us using invariant to flag for clearly invalid and irrecoverable cases such as a value missing from a map or other expectations that we have in the moment we're writing the code. This way, we can more confidently write convoluted control flow while getting context aware invariants.

My mentality is derived from [Matt Klein's article written while he worked on Envoy proxy](https://medium.com/@mattklein123/crash-early-and-crash-often-for-more-reliable-software-597738dd21c5) and [the Erlang approach](https://ferd.ca/the-zen-of-erlang.html) at large.

PR #211, opened 2022-06-10
Fix popover behavior when popover is open and the trigger is clicked

AndrewPrifer:
This PR fixes the popover flicker behavior when trying to close the popover through clicking on its trigger button by also excluding the trigger from the "outside" region.

What's missing still: some popovers aren't only closed on click outside, but also on blur, the use of which the Basic*Inputs specifically encourage for these purposes. The issue is that blur is also called on click, in which case we have the same close-open flicker as with onclickoutside. Would be nice to have a blur-like mechanism for closing popovers that only triggers on Enter and Escape. EDIT: For now the best option seems to be removing close on blur (inline keyframe editors already behaved like this, intentionally or not), and merge what we have. We can come up with a better solution than blur later.


theatre/studio/src/panels/DetailPanel/ProjectDetails.tsx
```diff
@@ -55,7 +55,7 @@ const ProjectDetails: React.FC<{
     }, 40000)
   }, [])
 
-  const [tooltip, openExportTooltip] = usePopover(
+  const {node: tooltip, open: openExportTooltip} = usePopover(
```
colelawrence, line 58: Suggestion: Rename these to not unstructured `const exportTooltip = usePopover(` and in call sites just use `exportTooltip.open(node, mouseEvent)` or whatever.

colelawrence (dismissed):

theatre/studio/src/panels/DetailPanel/ProjectDetails.tsx
AndrewPrifer: Oh I agree, I just didn't want the changelog to be a mess by also modifying call-sites of open/close etc

AndrewPrifer:
@colelawrence what do you think about the bit where we use blur events to notify popovers that they should close, but these blur events also trigger on click, which we already handle in the popover?

colelawrence:
@AndrewPrifer re: blur events, I'm not sure about UX. I do not feel strongly about blinking popover. I recommend merging whatever is valuable for now. And perhaps adding inline comments around the blur handling code about this weird UX.

Perhaps a good idea will come to us later on.

AndrewPrifer:
@colelawrence I agree, let's merge it now, the only issues are with popovers with inputs, like the end indicator and the playhead. Coincidentally, the close-on-blur mechanism was omitted from the inline keyframe editor.

theatre/studio/src/panels/SequenceEditorPanel/DopeSheet/Right/AggregatedKeyframeTrack/AggregateKeyframeEditor/AggregateKeyframeConnector.tsx
```diff
@@ -55,7 +55,11 @@ export const AggregateKeyframeConnector: React.VFC<IAggregateKeyframeConnectorPr
     const [contextMenu] = useConnectorContextMenu(props, node)
     const [isDragging] = useDragKeyframe(node, props.editorProps)
 
-    const [popoverNode, openPopover, closePopover] = usePopover(
+    const {
+      node: popoverNode,
+      toggle: togglePopover,
+      close: closePopover,
+    } = usePopover(
```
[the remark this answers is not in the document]
AndrewPrifer: Done

theatre/studio/src/panels/DetailPanel/ProjectDetails.tsx
AndrewPrifer: Done

PR #213, opened 2022-06-12
Replace Vite with ESBuild for the playground

AndrewPrifer:
This PR should reproduce exactly the setup we have now, but with pure ESBuild.

- [x] hot reload
- [x] e2e tests pass locally
- [x] e2e tests pass in CI
- [x] fast
- [x] Vercel deployment works (Vercel, for some reason, returns the root index.html for any url that doesn't have a dot in it, probably a bandaid for the previous build, since this new build works with a bare-bones static file server-FIXED)


colelawrence (changes requested):
`yarn playground` does not seem to serve anything on initial load after running `yarn install`.

Change: `build.tsx` can be just `build.ts`

AndrewPrifer:
@colelawrence can't reproduce it not working, any ideas what might go wrong?

As for build.tsx, it does have JSX in it.

AriaMinaei (approved):

colelawrence (approved):
Adds value. I'll come back soon with a PR for adjusting small things like auto port assignment.
Edit, that PR was #215

PR #216, opened 2022-06-13
Fix transparencies for browsers with no backdrop blur

AndrewPrifer:
As per Mariusz' updated designs for FF and friends.


colelawrence (approved):
LGTM

PR #218, opened 2022-06-13
dev: Initial playground re-design

colelawrence:
Cole's screen capture of Playground – Theatre js 2022-06-16 at 13 48 05@2x


packages/playground/devEnv/build.ts
```diff
@@ -3,294 +3,224 @@ import {writeFile, readFile} from 'fs/promises'
...
+      esbuildWatchStop = buildResult?.stop
+      // Read index.html template
+      const index = await readFile(path.join(__dirname, 'index.html'), 'utf8')
+      await Promise.all([
+        // Write home page
+        writeFile(
+          path.join(buildDir, 'index.html'),
+          index
+            .replace(/<\/head>/, `${homeHtml.head}<\/head>`)
+            .replace(/<body>/, `<body>${homeHtml.html}`),
```
AndrewPrifer, line 159: This keeps the %ENTRYPOINT% placeholder intact, resulting in an error 400. I deliberately replaced the whole body, since we don't want to put anything else there anyway.

packages/playground/devEnv/home/ItemSectionWithPreviews.tsx
```diff
@@ -0,0 +1,130 @@
...
+    height: var(--previewHeight);
+    width: var(--previewWidth);
+  }
+`
+const HomeItem = styled.a`
+  border: 1px solid rgba(255, 255, 255, 0.08);
+  border-radius: 4px;
+  text-decoration: none;
+  overflow: hidden;
+
```
vezwork, line 127: nit: add :hover style / :active style for keyboard nav

packages/playground/devEnv/build.ts
colelawrence: Thanks for pointing this out!

vezwork (approved):

vezwork:
No complaints, LGTM!

packages/playground/devEnv/home/ItemSectionWithPreviews.tsx
colelawrence: I don't think this actually makes a difference AFAICT.

If you're using Firefox on macOS like me, you may have to press Opt+F7 to switch your macOS leve tabbing behavior to get normal seeming tabbing.

PR #219, opened 2022-06-13
Add thumbnail generation for playground (PoC)

AndrewPrifer:
This PR demonstrates how to generate thumbnails for the playground pages.

Generation happens if either
- the thumbnails directory doesn't yet exist
- the `-t` flag is supplied

Thumbnail generation doesn't block the dev server from starting and serving requests, since the operations are asynchronous and most of the work happens outside of Node.

Since this relies on starting the server, to make this work for regular builds that don't need the dev server running, we could just run the server while generating the thumbnails and then shut it down.

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,6 +56,7 @@
     "lint-staged": "^11.1.2",
     "node-gyp": "^8.1.0",
     "prettier": "^2.3.2",
+    "puppeteer": "^14.3.0",
     "typescript": "^4.4.2",
     "zx": "^2.0.0"
   },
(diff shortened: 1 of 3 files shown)


colelawrence (changes requested):
I appreciate this code being written and can see a future where it makes sense to include something of this sort, but in its current state, I'm apprehensive about including puppeteer as a dev dependency and potentially extending the time CI takes by any measure unless strictly necessary.

AndrewPrifer:
Closing this PR now since it served its use.

PR #224, opened 2022-06-15
dev: Build playground files much faster and add watch

colelawrence:
I think this PR has legs, and I feel pretty good about how the changes are laid out. I'd like to spend a few weeks with these changes in the codebase, before committing to re-working other parts of our build process to also use this approach.

### DIFF
--- a/packages/playground/devEnv/.gitignore
+++ b/packages/playground/devEnv/.gitignore
@@ -1,0 +1,1 @@
+build.compiled.js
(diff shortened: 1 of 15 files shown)


colelawrence:
I think this PR has legs, and I feel pretty good about how the changes are laid out. I'd like to spend a few weeks with these changes in the codebase, before committing to re-working other parts of our build process to also use this approach.

Marking as ready for review.

AriaMinaei (approved):

PR #231, opened 2022-06-22
Fix playground nav to examples in localhost

colelawrence:
Please test locally and let me know if this fixes the issues experienced!


vezwork (approved):
Works great. LGreatTM

PR #232, opened 2022-06-24
Assorted fixes and minor smoothing for using playground

colelawrence:
Internally tracked via Notion https://www.notion.so/theatrejs/Shallower-on-ramp-2022-06-9eb5d5ddfb874f74b5575

https://user-images.githubusercontent.com/2925395/175536616-1de60045-192f-4f70-b358-f1fd016a9f2e.mp4

8b0d359bf78

### DIFF
--- a/packages/playground/README.md
+++ b/packages/playground/README.md
@@ -1,6 +1,6 @@
 # The playground
 
-The playground is the quickest way to hack on the internals of Theatre. It also hosts our end-to-end tests. It uses a build setup that builds all the related packages in one go, so you _don't_ have to run a bunch of build commands separately to start developing.
+The playground is the quickest way to hack on the internals of Theatre. It also hosts our end-to-end tests. It uses a build setup (see the live-reload esbuild server in [./devEnv/build.ts](./devEnv/build.ts)) that builds all the related packages in one go, so you _don't_ have to run a bunch of build commands separately to start developing.
 
 ## Directory structure
 
--- a/packages/playground/README.md
+++ b/packages/playground/README.md
@@ -1,6 +1,6 @@
 # The playground
 
-The playground is the quickest way to hack on the internals of Theatre. It also hosts our end-to-end tests. It uses a simple vite config that builds all the related packages in one go, so you _don't_ have to run a bunch of build commands separately to start developing.
+The playground is the quickest way to hack on the internals of Theatre. It also hosts our end-to-end tests. It uses a build setup that builds all the related packages in one go, so you _don't_ have to run a bunch of build commands separately to start developing.
 
 ## Directory structure
 
--- a/packages/playground/devEnv/build.ts
+++ b/packages/playground/devEnv/build.ts
@@ -1,4 +1,4 @@
-import {readdirSync} from 'fs'
+import {readdirSync, readFileSync, statSync} from 'fs'
 import {writeFile, readFile} from 'fs/promises'
 import path from 'path'
 import type {BuildOptions} from 'esbuild'
@@ -18,6 +18,7 @@ import {createServerForceClose} from './createServerForceClose'
 
 const playgroundDir = (folder: string) => path.join(__dirname, '..', folder)
 const buildDir = playgroundDir('build')
+const srcDir = playgroundDir('src')
 const sharedDir = playgroundDir('src/shared')
 const personalDir = playgroundDir('src/personal')
 const testDir = playgroundDir('src/tests')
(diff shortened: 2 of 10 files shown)


theatre/shared/src/propTypes/utils.ts
```diff
@@ -8,7 +8,9 @@ import type {
 import type {PathToProp} from '@theatre/shared/utils/addresses'
 import type {$IntentionalAny} from '@theatre/shared/utils/types'
 
-export function isPropConfigComposite(
+/** @deprecated renamed to {@link isPropConfigCompound} for clarity */
+export const isPropConfigComposite = isPropConfigCompound
+export function isPropConfigCompound(
```
AriaMinaei, line 13: This one is undocumented, but enums would be composite but not compound.

packages/playground/devEnv/build.ts
```diff
@@ -51,17 +55,45 @@ export async function start(options: {
...
-                {
-                  entryDir: path.join(groupDir, moduleDirName),
-                  outDir: path.join(
-                    buildDir,
-                    path.basename(groupDir),
-                    moduleDirName,
-                  ),
-                },
-              ]),
+              readdirSync(groupDir)
```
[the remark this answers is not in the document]
colelawrence: This is still for internal. I wrote this before really considering the split between end-user building and internal building. Since it was already written (and it's fairly straightforward and it could reduce friction for internal dev), I opted to keep it.

When picking an end-user tool for playing with Theatre, I'd certainly make sure this just happens automatically via the bundler we choose.

theatre/shared/src/propTypes/utils.ts
colelawrence: Okay, I will add a comment and drop this commit.

colelawrence: Applied

packages/playground/src/shared/dom/Scene.tsx
```diff
@@ -76,8 +76,18 @@ const Box: React.FC<{
   sheet: ISheet
   selection: IStudio['selection']
 }> = ({id, sheet, selection}) => {
+  const defaultConfig = useMemo(
+    () =>
+      Object.assign({}, boxObjectConfig, {
+        // give the box initial values offset from each other
+        x: ((id.codePointAt(0) ?? 0) % 15) * 100,
```
vezwork, line 83: I understand this results in offset boxes, not sure what this is doing though. Is it like a random number generator? why `id.codePointAt(0)`? 

Oh I think I see, the ids are numbers that are increasing so this works like that? 

Break out into function?

theatre/shared/src/utils/color.ts
```diff
@@ -42,12 +42,20 @@ export function parseRgbaFromHex(rgba: string) {
...
+  } = {},
+) {
+  const alpha = ((rgba.a * 255) | (1 << 8)).toString(16).slice(1)
+
   const hex =
     ((rgba.r * 255) | (1 << 8)).toString(16).slice(1) +
     ((rgba.g * 255) | (1 << 8)).toString(16).slice(1) +
     ((rgba.b * 255) | (1 << 8)).toString(16).slice(1) +
-    ((rgba.a * 255) | (1 << 8)).toString(16).slice(1)
+    (removeAlphaIfOpaque && alpha === 'ff' ? '' : alpha)
```
vezwork, line 58: nice

packages/playground/README.md
```diff
@@ -1,6 +1,6 @@
 # The playground
 
-The playground is the quickest way to hack on the internals of Theatre. It also hosts our end-to-end tests. It uses a simple vite config that builds all the related packages in one go, so you _don't_ have to run a bunch of build commands separately to start developing.
+The playground is the quickest way to hack on the internals of Theatre. It also hosts our end-to-end tests. It uses a build setup that builds all the related packages in one go, so you _don't_ have to run a bunch of build commands separately to start developing.
```
vezwork, line 4: This change makes the doc slightly less descriptive, but the build setup seems to have increased in complexity. Perhaps you can add an extra sentence or two to describe the build setup/stack/process?

vezwork (approved):
Approved with readability suggestions

packages/playground/src/shared/dom/Scene.tsx
colelawrence: I don't really care enough to try to pull it out. Essentially, it just uses the first char of the string as a reference of code point (e.g. `"A".charCodeAt(0)` -> `65`) then uses that as a modulo'd number with 100 px offset. It's dirty, but I don't care too much since the comment is there. To break it out puts a special kind of pressure on the code being clean / re-usable which it absolutely isn't 😂

vezwork: fair enough

PR #233, opened 2022-06-24
feature: Add control indicators for Compounds in SequenceEditor

colelawrence:
Internally written up at [Add compound keyframe addition via sequence editor (add keyframe cursors to compound sheet row)](https://www.notion.so/theatrejs/Add-compound-keyframe-addition-via-sequence-editor-add-keyframe-cursors-to-compound-sheet-row-c02d0fb0cfd54fc9a9216381f7ad88b0)


vezwork (approved):
Code LGTM. Product-wise I think its decent enough to merge and come back to. Its slightly strange that the buttons are to the left of the name as opposed to to the right on the leaf props, but it maintains the visual hierarchy which I like. Would defer comments/further changes to others trying it out once its merged tho.

colelawrence:
Added Aria to review to verify that it's okay to start with the UI shown in screenshot.
The other caveat is that the aggregate control is not shown at the whole SheetObject row level.

colelawrence:
Now waiting on design to confirm what to do with right aligning styles from Mariusz

image

PR #240, opened 2022-06-29
Enable directly using custom editable elements through the editable object

AndrewPrifer:


AndrewPrifer:
@donmccurdy could you please take a look at this and see if it solves your problem re: custom editables without having to mess with the schema?

AndrewPrifer:
@donmccurdy there are indeed a couple easy ways already to accommodate custom threejs objects. `e.primitive` would be one, or if you want to control how theatre treats these objects, the following was already possible: `extend({ MyMesh }); const MyEditableMesh = editable('myMesh', 'mesh');`, I just needed to accommodate it in typescript as well (see https://github.com/theatre-js/theatre/pull/237).

The rationale for this PR was that from our talks it seemed like the requirement was not to _just_ make this possible, but specifically to make it seamless. @juniorxsound @donmccurdy can you think of something that could better accommodate your use-case in this sense?

I do like this implementation because while it does slightly extend the API surface, it is fully backwards-compatible, would be opt-in, and for users who are fully committed to r3f and would use `extend({...})` anyway, it is completely free as soon as I get r3f to expose the `ThreeElements` type (which would be super useful anyway not just for us, but for react-spring as well, whose types are completely broken right now because of the type complexity `React.ElementType` introduces).

Unless you think there could be other solutions here with better ergonomics (which I'm specifically targeting as my goal), I'd be inclined to merge this.

As for coining the scene, yes, the only requirement should be that you implement `.clone()`, we even peer-depend on threejs so that nothing breaks in the transition between your setup and the editor.

donmccurdy:
I'll defer to @juniorxsound on the remaining questions of API surface. As background, our use cases involve writing a limited amount of React / R3F code up front for tooling that is reused by many more end-users who will create three.js objects without interacting with the React codebase at all. So IMHO we can afford to declare the theatre/studio sheet object properties exhaustively ourselves, seamlessness there is not critical. Whereas doing something reasonable with current and future 'new' three.js objects (even if it's just exposing pos/rot/scale) without breaking is crucial.

> As for coining the scene, yes, the only requirement should be that you implement .clone(), we even peer-depend on threejs so that nothing breaks in the transition between your setup and the editor.

I have to admit I'd prefer that theatre/r3f append a managed `THREE.Group` in its own layer to the existing scene, containing 3D UI elements and helpers for the snapshot view, rather than cloning a potentially large scene. Realtime sync would be easy in that case. But I'm off topic for the thread and this isn't currently a blocker for us. :)

juniorxsound:
Hi 👋 sorry for jumping into the discussion late.

Echoing @donmccurdy I think smaller API surface might be a beneficial thing overall (and for what we need, the `<e.primitive>` approach seems to be working well). Two potential use cases I do see for being able to extend the editable type is
- [ ] Being able to add a custom helper to an editable object in the snapshot viewer
- [ ] Being able to add an icon (similar to the camera icon) for invisible objects in the snapshot viewer

This is a little bit of a tangent to what's proposed above but might be worth considering?

AndrewPrifer:
@juniorxsound all of what you mention is going to be enabled by #166. This PR here is only for convenience. From our discussions I gathered that the idea was to enable seamless use of custom objects without any upfront configuration, hence this PR, but since you don't seem to need it right now, I'm going to sit on it a bit more. :)

@donmccurdy one of the first things on my plate is going to be a real-time architecture, I completely agree that it would bring massive value.

Issue #261, opened 2022-07-20
Many event listeners created by theatre/studio, possible memory leak


donmccurdy:
We've been noticing that while scrubbing the sequence position, theatre/studio appears to create a lot (thousands?) of listeners, and the JS Heap climbs pretty fast as well. These eventually do get cleaned up, but it seems like a possible performance bottleneck that we weren't seeing before the v0.5.x prereleases. We encountered this in our own project but it's easily reproducible by scrubbing the sequence time in a [playground link](https://theatre-playground-theatrejs.vercel.app/shared/r3f-rocket) as well. In the screenshot below, each area where the graph is climbing corresponds to when I was scrubbing the sequence. I see the same issue when driving the sequence by scrolling in our application, but I don't see the issue if theatre/studio is not hooked up:

<img width="952" alt="Screen Shot 2022-07-20 at 1 28 20 PM" src="https://user-images.githubusercontent.com/1848368/180055968-e5b74473-b534-4fe4-9391-dccbf6c6cd68.png">

AriaMinaei:
Thanks @donmccurdy! This is a pretty urgent issue. It'll be fixed in the next release!

vezwork:
Hey @donmccurdy are you still facing this issue or was it fixed in 0.5?

PR #269, opened 2022-07-26
perf: Initial hand optimization of ControlIndicators

colelawrence:
Related to #261
This intends to begin addressing issues with changing sequence position at 60fps
https://linear.app/theatre/issue/P-268/fix-the-memory-leak


colelawrence:
Closing in favor of #271

PR #273, opened 2022-08-03
Make derivation keepHot when using obj.value P-217

colelawrence:
# Keep commits when merging

See https://linear.app/theatre/issue/P-217/if-objvalue-is-read-make-sure-i

### DIFF
--- a/packages/playground/src/tests/reading-obj-value/index.tsx
+++ b/packages/playground/src/tests/reading-obj-value/index.tsx
@@ -1,0 +1,66 @@
+import studio from '@theatre/studio'
+import {getProject, types} from '@theatre/core'
+import state from './reading obj value.theatre-project-state.json'
+studio.initialize({usePersistentStorage: false})
+
+const project = getProject('reading obj value', {state})
+const TOTAL_ELEMENTS = 300
+const TOTAL_ELEMENTS_R = 1 / TOTAL_ELEMENTS
+const playbackControlObj = project
+  .sheet('controls sheet')
+  .object('playback control', {
+    interval: types.number(50, {
+      range: [1, 20000],
+    }),
+  })
+const elements = new Array(TOTAL_ELEMENTS).fill(0).map((_, idx) => {
+  const sheet = project.sheet('sample sheet', `#${idx}`)
+  const obj = sheet.object('sample object', {
+    position: {
+      x: 0,
+      y: 0,
+    },
+  })
+
+  const el = document.createElement('div')
+
+  Object.assign(el.style, {
+    position: 'absolute',
+    height: '2rem',
+    width: '2rem',
+    borderRadius: '50%',
+    background: `hsl(${idx * 360 * TOTAL_ELEMENTS_R}, 100%, 80%)`,
+    left: 'var(--x)',
+    top: 'var(--y)',
+  })
+
+  document.body.appendChild(el)
+
+  return {el, sheet, obj}
+})
+
+project.ready.then(() => {
+  // select the playback controls obj so it shows as a tweakable control
+  studio.setSelection([playbackControlObj])
+  for (let i = 0; i < elements.length; i++) {
+    const sheet = elements[i].sheet
+    sheet.sequence.position = i * TOTAL_ELEMENTS_R * 5
+    sheet.sequence.play({
+      iterationCount: Infinity,
+    })
+  }
+})
+
+function render() {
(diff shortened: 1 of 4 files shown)


vezwork (changes requested):

theatre/core/src/sheetObjects/TheatreSheetObject.ts
```diff
@@ -153,8 +160,27 @@ export default class TheatreSheetObject<
...
+        }
+        this._keepHotSchedule = {
+          schedule: debounce(clear, KEEP_HOT_FOR_MS),
+          clear,
+        }
+      }
+
+      this._keepHotSchedule?.schedule()
+    }
+    return der.getValue()
```
vezwork, line 183: I rewrote this function to understand it, came up with this equivalent(?) code which has a few less moving parts:

```suggestion
private _keepHotDebouncedUntap: undefined | VoidFn
// ...
 const der = this._valuesDerivation()
    if (KEEP_HOT_FOR_MS != null) {
      if (!der.isHot && this._keepHotDebouncedUntap == null) {
        const untap = der.keepHot()
        this._keepHotDebouncedUntap = debounce(untap, KEEP_HOT_FOR_MS)
      }

      this._keepHotDebouncedUntap?.()
    }
    return der.getValue()
```
```

File: theatre/core/src/sheetObjects/TheatreSheetObject.ts [hunk 2]
```diff
@@ -153,8 +160,27 @@ export default class TheatreSheetObject<
...
 
+  // internal: Make the deviration keepHot if directly read
   get value(): PropsValue<Props> {
-    return this._valuesDerivation().getValue()
+    const der = this._valuesDerivation()
+    if (KEEP_HOT_FOR_MS != null) {
+      if (!der.isHot && this._keepHotSchedule == null) {
+        const untap = der.keepHot()
+        const clear = () => {
+          if (this._keepHotSchedule?.clear === clear) {
```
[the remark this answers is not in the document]
colelawrence: Not necessarily, because the `_keepHotSchedule` can be re-assigned by a different accessor before `_keepHotSchedule.clear()` is called ()or is something like `setTimeout(_keepHotSchedule.clear, 2000)` is used.

Any ideas for improvement of how I did this are welcome, though.

theatre/core/src/sheetObjects/TheatreSheetObject.ts
colelawrence: I'll have to look at this closer. Thanks for the suggestion. I definitely didn't love the feeling of how this was written before. It does seem simpler to just use a debounce helper

theatre/core/src/sheetObjects/TheatreSheetObject.ts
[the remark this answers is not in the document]
vezwork: Ah right, thank you for explanation. I think it may be able to simplified (see my other comment). Slightly tricky to reason about this code since the order of execution is not straightforward.

packages/playground/src/tests/reading-obj-value/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+project.ready.then(() => {
+  for (let i = 0; i < elements.length; i++) {
+    const sheet = elements[i].sheet
+    sheet.sequence.position = i * TOTAL_ELEMENTS_R * 5
+    sheet.sequence.play({
+      iterationCount: Infinity,
+      // rate:  * 1.5 + 0.5,
+    })
+  }
+  // studio.setSelection([obj])
```
fulopkovacs, line 51: Do we want to leave comments in the playground files? Or a better question: are these lines worth being kept around (do we want to uncomment them later)?

fulopkovacs (approved):
I don't really have anything to add @vezwork 's review. I agree with the one modification he suggested (although I haven't tested it), other than that it's an LGTM from me.

theatre/core/src/sheetObjects/TheatreSheetObject.ts
colelawrence: LMK what you think of the new approach

packages/playground/src/tests/reading-obj-value/index.tsx
colelawrence: Applied

vezwork (approved):
LGTM. A bit hard to think through the keepHot code since it involves async code paths but it appears safe and correct.

PR #275, opened 2022-08-03
Add r3f stress + Ticker.__ticks counter

colelawrence:
Here, we're adding the r3f stress test playground item and augmenting Ticker to count up every time `tick()` is called for performance measurement.


packages/playground/src/tests/r3f-stress-test/SpaceStress.theatre-project-state.json
```diff
@@ -0,0 +1,1155 @@
+{
```
fulopkovacs, line 1: I'm not sure if it matters, but the rocket's animation is a bit off (the parts of the rocket are not moving together). If we won't share this example with people from our community that much, then it won't be a problem, but I just thought I'll point this out.

https://user-images.githubusercontent.com/43729152/182893243-d8297aff-db97-4177-a4ac-73f6b7dc6b04.mp4

fulopkovacs (approved):
LGTM. I left two comments, but feel free to leave the code unchanged after reading them.

vezwork (approved):
pair reviewed with Cole and LGTM!

packages/playground/src/tests/r3f-stress-test/index.tsx
```diff
@@ -0,0 +1,39 @@
...
+studio.extend(extension)
+studio.initialize()
+
+ReactDOM.render(<App />, document.getElementById('root'))
+
+const raf = Ticker.raf
+
+setInterval(() => {
+  const id = Date.now().toString(36) + '-'
+  performance.mark(id)
```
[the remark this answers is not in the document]
colelawrence: performance.now is more computationally expensive in V8, and it's a lot higher resolution than necessary because we end up just rounding the ticks per second.

packages/playground/src/tests/r3f-stress-test/SpaceStress.theatre-project-state.json
colelawrence: Yep, this has always been an issue. I'd love to see this glb file was fixed for both examples.

PR #277, opened 2022-08-05
refactor: Playground items now don't load studio in iframes

colelawrence:
* Greatly improves the load time of the playground

https://user-images.githubusercontent.com/2925395/183131403-927f8e2e-576d-44b5-9f0f-caa087aed16a.mp4

### DIFF
--- a/packages/playground/src/common/LazyLoadStudio.tsx
+++ b/packages/playground/src/common/LazyLoadStudio.tsx
@@ -1,0 +1,96 @@
+import React, {useEffect, useState} from 'react'
+
+enum ReadyState {
+  Errored,
+  Loaded,
+  Loading,
+  ShouldLoad,
+  AskToLoad,
+}
+
+type ImportWithArgs<A extends any[]> = Readonly<
+  [() => Promise<{loadStudio(...args: A): void | PromiseLike<void>}>, ...A]
+>
+
+/**
+ * @example
+ * <LazyLoadStudio load={() => import('./loadStudio')} />
+ *
+ * @example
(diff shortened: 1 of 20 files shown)


colelawrence:
# Squash and merge is good

packages/playground/src/common/LazyLoadStudio.tsx
```diff
@@ -0,0 +1,99 @@
...
+  const [state, setState] = useState(() =>
+    // don't auto load if in iframe
+    props.autoLoad && window.parent === window
+      ? ReadyState.ShouldLoad
+      : ReadyState.AskToLoad,
+  )
+  useEffect(() => {
+    if (state === ReadyState.ShouldLoad) {
+      setState(ReadyState.Loading)
+      const loadPromise =
```
vezwork, line 49: See my readability suggestion for the type of `import` above. Also see suggestion to make the scope async so await case be used. This suggestion would simplify this into a single case (ternary operator not necessary) and remove the need for nested callbacks.

packages/playground/src/common/LazyLoadStudio.tsx
```diff
@@ -0,0 +1,99 @@
+import React, {useEffect, useState} from 'react'
+
+enum ReadyState {
+  Errored,
+  Loaded,
+  Loading,
+  ShouldLoad,
+  AskToLoad,
+}
```
vezwork, line 9: I like this. Made it easy to think thru the async code.

vezwork (changes requested):

PR #280, opened 2022-08-16
Fix issue window undefined error in bundled studio

vezwork:
Issue: previously the bundled studio index.js included a bundled version of styled-components. Here's a line from the bundled code saying what version of styled components is included: `// ../node_modules/styled-components/dist/styled-components.browser.esm.js` (path relative to theatre/studio) the styled- components bundled code contains references to the window seemingly without checking that it is undefined. This is fixed by wrapping our UI import in a condition that checks if window is undefined, letting our bundler create bundled code that only runs styled-components if window is defined.


vezwork:
still testing and finding bugs, will merge this after some more testing tomorrow.

PR #284, opened 2022-08-23
Add Sheet aggregate track

vezwork:
Adds a sheet aggregate track above all the object tracks.

Screen Shot 2022-08-23 at 3 40 12 PM

Two ~un~resolved related bugs were found:
- ~keyframe inline editor popovers don't properly sanitize values before displaying them~
- ~aggregate track context-menu options get stuck thinking there is a selection even after the selection is cleared.~


vezwork:
Just noticed a behaviour I did not intend. Copying an object's aggregate keyframe of a single primitive keyframe does not seem to correctly set the path of the keyframe: pasting back into the aggregate track pastes into all sub-tracks instead of just the track of the single primitive keyframe that was aggregated and copied.

PR #285, opened 2022-08-24
Replace uniqueName with theatreKey

AndrewPrifer:
In case you have competing ideas for the name, I am fairly adamant on these constraints:
- has to include the word "theatre"
- has to be reasonably short
- has to include the word "id", "key", "name", or something similar denoting an identifier


vezwork:
I'm happy with the name. "TheatreObjectKey" would've been my first pick but its too long. LGTM

vezwork (approved):

PR #286, opened 2022-08-25
Require Opt modifier for orbiting

AndrewPrifer:
The operations accessible through the left mouse button are getting a little crowded.


vezwork:
@AndrewPrifer let me know if there's anything you'd like me to specifically take a look at? Otherwise, I see that this is a fork of the orbit controls. It seems good to me, we were going to have to fork it at some point or develop our own down the line. I assume the modification to change the controls was straightforward, if so LGTM from me.

PR #309, opened 2022-10-07
Apply common sense constraints to window drag and resize

AndrewPrifer:
This PR fixes the issue that extension windows can be dragged/resized in such a way that the drag area no longer falls inside the browser viewport and the window becomes impossible to move.


vezwork (approved):
Could also add these changes to the dope sheet.

AndrewPrifer:
Dope sheet is fixed. Problem was that while the left side uses our regular PanelDragZone component, on the right side the focus range doubles for panel dragging too, and I didn't apply the constraints there.

PR #320, opened 2022-10-20
Create a notification system that can display notifications in Theatre.js' Studio

AndrewPrifer:
The goal of this PR is to create a unified notification system that can be used across core, studio, and extensions.

Goals:
- Make it easy to create high quality notifications that provide helpful information in an easy to parse way.
- Should be a unified system for displaying both Studio notifications, and console notifications
- Should be unobtrusive, but unmissable

The way this is achieved is the following.

The notification API provides a structure that makes it easy to craft high-quality notifications:

```ts
notify.warning(
  // Title
  'Invalid path provided to object',
  // Message. Can include markdown.
  `The path in \`${fnName}("${unsanitisedPath}")\` was sanitized to \`"${sanitisedPath}"\`.\n\n` +
    'Please replace the path with the sanitized one, otherwise it will likely break in the future.',
  // Relevant docs
  [
    {
      url: 'https://www.theatrejs.com/docs/latest/manual/objects#creating-sheet-objects',
      title: 'Sheet Objects',
    },
    {
      url: 'https://www.theatrejs.com/docs/latest/api/core#sheet.object',
      title: 'API',
    },
  ],
)
```

This will cause a notification to appear in studio's notification manager:
Playground – Theatre js 2022-10-20T13 08 17@2x

Notifications can freely include markdown paragraphs, code blocks, lists, etc.

Additionally, for warnings, a console log is also printed with the same message.

The collapsed notification manager shows the types of notifications currently in the list:
Playground – Theatre js 2022-10-20T13 11 17@2x

Notifications are by default deduplicated. A notification is only shown if there isn't already one displayed with the same content. This is useful since while we often don't care about cluttering the console, the studio UI is a different matter. Calling `notify` should be idempotent in most cases. This way we don't have to worry about react rendering twice in a row, or calling `notify` in a loop even.

You can opt out of this behavior through the last argument:

```ts
// this notification will be fired as many times as this code is run
notify.warning(
  'My notification',
  'Some message',
  [],
  // allow duplicates
  true
)
```

The notification system is accessible across studio, extensions, and core.

When used in studio, the API is accessible through the following import:

```ts
import {notify} from '@theatre/studio/notify'
```

When used in core, the API is accessible through the following import:

```ts
import {notify} from '@theatre/shared/notify'
```

When used in extension code, the API is accessible through the following import:

```ts
import {notify} from '@theatre/studio'
```

Note, notifications fired in core will only be displayed if studio is loaded.


AndrewPrifer:
@vezwork I think windows also has overlay scrollbars now. There's not much we can do about them for now, scrollbar styling is barely supported, and what is supported mostly isn't on standards track. In a few years. Or by using virtual scrolling, but then we should do them everywhere, the outline is a similarly ugly offender

theatre/shared/src/notify.ts
```diff
@@ -0,0 +1,57 @@
+import logger from './logger'
```
vezwork, line 1: Aria and I discussed deprecating / removing usages of this logger previously (because of the implementation is complex and most usages of it require a logger instance which is hard to maintain). Could be replaced with a console.log?

packages/playground/src/shared/notifications/Scene.tsx
```diff
@@ -0,0 +1,232 @@
+import studio from '@theatre/studio'
```
vezwork, line 1: Suggestion: simplify this playground example so it has nothing in it except the notifications code. Would help me in the future if I'm looking to reference notification usage and don't have to look through the playground example as much to find it.

vezwork (approved):
Sweet

theatre/shared/src/notify.ts
AndrewPrifer: I'd say let's treat the deprecation of the logger in a future PR, this one should be focused on notifications. Usages should be easy to find and replace, so I'd keep things consistent for now and keep using logger here.

packages/playground/src/shared/notifications/Scene.tsx
AndrewPrifer: Alright, I'll simplify it a bit

theatre/shared/src/notify.ts
vezwork: Makes sense to me

packages/playground/src/shared/notifications/Scene.tsx
vezwork: 👍🙂

AndrewPrifer: Done. I kept the background color code because it helps us easily catch UI issues.

AndrewPrifer:
I'll go ahead and merge this since it is easy to revert/modify later. The biggest ambiguity is how we should expose `notify` from studio (right now it's just exported as its own symbol), and how we should expose it to core (right now it's just attached to `window`). Both can be easily changed later, and this way we can start using the API. If for whatever reason the UI doesn't work and we want to quickly get rid of it, that's also fine, since the `notify` API also takes care of logging to the console, so we won't suddenly have to go back and revert all those changes back to logger.

theatre/shared/src/notify.ts
```diff
@@ -47,7 +47,7 @@ const createHandler =
     }
 
     // @ts-ignore
-    return window[globalVariableNames.notifications]?.notify[type](...args)
+    return window?.[globalVariableNames.notifications]?.notify[type](...args)
```
AndrewPrifer, line 50: @vezwork our policy now is to treat all @theatre/core  usages of `window` as optional, right? I don't quite remember the work you did on keeping compatible with nextjs.

PR #321, opened 2022-10-20
Experiment: Generator function API for prism (derive)

vezwork:
also change IDerivation methods:
- deprecate `changesWithoutValues`
- add `onStaleWithoutValues`
- add `onStale`

TODO: make code exmaple of `derive`

Open questions:
For the PR:
- how to replicate prism.memo, prism.effect, etc. using only derivations?
- can `derive` be more performant than `prism`? i.e. does using generator functions give an edge over using `discoveryMechansim`?

For myself:
- do atoms notify if references inside them change? Or only if values change?
- Can you use val inside of prism effect and memo? (I am almost sure but need to test)
- what is a good/simple mental model/heuristic for derivations? I still have trouble keeping track of hotness and freshness and what to worry about with them.


AriaMinaei:
Closing in favor of #345

PR #329, opened 2022-10-28
Add `nudgeIgnoreRange` opt to number type

### LINKED ISSUE #314: [Bug] - Type:Number - Multiple inputs break
Using `nudgeMultiplier` and `range` currently only uses `range`.

```
import { types as t } from '@theatre/core'

sheet
  .object('Example', {
    multiplierWorks: t.number(0, { nudgeMultiplier: 1 }),
    rangeWorks: t.number(0, { range: [-1, 100] }),
    nudgeError: t.number(0, { nudgeMultiplier: 1, range: [-1, 100] }),
  })
  .onValuesChange((values: any) => {
    console.log(values)
  })
```

vezwork:
fixes #314 

@tomorrowevening and @clementroche were experiencing an issue (#314) where using `types.number` with `nudgeMultiplier` and `range` at the same time is behaving unexpectedly. I understand that the source of the unexpected behaviour is that the nudge is automatically scaled to the size of the `range`, which can make the nudgeMultiplier behave very differently for different `range`s. To fix this issue, I have added an additional `types.number` option `nudgeIgnoreRange` to disable this behaviour so that `nudgeMultiplier` results in the same nudge size regardless of the `range`.

### DIFF
--- a/packages/playground/src/shared/turtle/TurtleRenderer.tsx
+++ b/packages/playground/src/shared/turtle/TurtleRenderer.tsx
@@ -17,7 +17,11 @@ studio.initialize()
 
 const objConfig = {
   startingPoint: {
-    x: types.number(0.5, {range: [0, 1]}),
+    x: types.number(0.5, {
+      range: [-1, 1],
+      nudgeMultiplier: 0.1,
+      nudgeIgnoreRange: true,
+    }),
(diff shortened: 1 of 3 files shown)


AriaMinaei:
Closing as fixed

PR #332, opened 2022-11-03
document and simplify cache usage, SheetObject

vezwork:
I was having quite a bit of trouble understanding the usages of the `SimpleCache` and `prism.memo` in `SheetObject.ts` and `SheetObjectTemplate.ts`. I have refactored to make the cache usage more explicit (lazy initialization) and documented. I have changed some `get___(): IDerivation<___>` methods into property getters to simplify code and help communicate that this is a single derivation and not getting a new derivation every time it is called/accessed. I also factored out a lot of logic in `getSequencedValues`.

### DIFF
--- a/packages/dataverse/src/Atom.ts
+++ b/packages/dataverse/src/Atom.ts
@@ -19,22 +19,18 @@ enum ValueTypes {
   Other,
 }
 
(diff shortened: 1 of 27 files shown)


AriaMinaei:
Closing in favor of the upcoming expressions feature

PR #363, opened 2022-12-22
Create idb-backed image prop

AndrewPrifer:


theatre/core/src/projects/DefaultAssetStorage.ts
```diff
@@ -0,0 +1,172 @@
...
+  project: Project
+  baseUrl?: string
+}): IAssetStorageConfig => {
+  return {
+    coreAssetStorage: {
+      getAssetUrl: (assetId: string) => `${baseUrl}/${assetId}`,
+    },
+    createStudioAssetStorage: async () => {
+      // Check for support.
+      if (!('indexedDB' in window)) {
```
AriaMinaei, line 41: might break ssr

theatre/core/src/projects/DefaultAssetStorage.ts
```diff
@@ -0,0 +1,172 @@
...
+  const keyframeValues = sheets
+    .flatMap((sheet) => Object.values(sheet?.sequence?.tracksByObject ?? {}))
+    .flatMap((tracks) => Object.values(tracks?.trackData ?? {}))
+    .flatMap((track) => track?.keyframes)
+    .map((keyframe) => keyframe?.value)
+
+  const allValues = [...staticValues, ...keyframeValues].filter(
+    // value is string, and is unique
+    (value, index, self) =>
+      typeof value === 'string' && self.indexOf(value) === index,
```
AriaMinaei, line 22: how does this filter over `types.string()` values?

theatre/core/src/projects/DefaultAssetStorage.ts
```diff
@@ -0,0 +1,172 @@
...
+          }
+        }),
+      )
+
+      // Clean up idb entries exported to disk
+      await Promise.all(
+        idbKeys.map(async (key) => {
+          const assetUrl = `${baseUrl}/${key}`
+
+          const response = await fetch(assetUrl, {method: 'HEAD'})
```
AriaMinaei, line 77: should `try{...}`, to catch exceptions like CORS errors

theatre/core/src/projects/DefaultAssetStorage.ts
AndrewPrifer: Hmm does SSR wait for async code? This does not run until studio is initialized. But yeah maybe better to be defensive about it. 😅

theatre/core/src/projects/DefaultAssetStorage.ts
```diff
@@ -0,0 +1,172 @@
...
+        },
+        createAsset: async (asset: File) => {
+          const existingIDs = getAllPossibleAssetIDs(project)
+
+          let sameSame = false
+
+          if (existingIDs.includes(asset.name)) {
+            const existingAsset =
+              assetsMap.get(asset.name) ??
+              (await fetch(`${baseUrl}/${asset.name}`).then((r) =>
```
AriaMinaei, line 121: also a `try{}` here

theatre/core/src/projects/Project.ts
```diff
@@ -172,11 +219,18 @@ export default class Project {
   }
 
   get ready() {
-    return this._readyDeferred.promise
+    return Promise.all([
```
AriaMinaei, line 222: This will re-create a promise for each `.ready` call. Let's cache that promise

theatre/core/src/propTypes/index.ts
```diff
@@ -690,6 +729,9 @@ export interface PropTypeConfig_StringLiteral<T extends string>
 
 export interface PropTypeConfig_Rgba extends ISimplePropType<'rgba', Rgba> {}
 
+export interface PropTypeConfig_Image
+  extends ISimplePropType<'image', string> {}
```
AriaMinaei, line 733: It'll be hard to distinguish between image and a simple string. Use `{type: 'Image', path: string}` instead

theatre/core/src/propTypes/index.ts
```diff
@@ -138,6 +138,45 @@ export const compound = <Props extends UnknownShorthandCompoundProps>(
...
+  } = {},
+): PropTypeConfig_Image => {
+  if (process.env.NODE_ENV !== 'production') {
+    validateCommonOpts('t.image(defaultValue, opts)', opts)
+  }
+
+  return {
+    type: 'image',
+    // FIXME
+    default: undefined as $FixMe,
```
AriaMinaei, line 171: `default` value cannot be unassignable to `valueType`.

packages/playground/src/shared/image/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+import {getProject, types} from '@theatre/core'
+import studio from '@theatre/studio'
+import React, {useEffect, useState} from 'react'
+import {render} from 'react-dom'
+import styled from 'styled-components'
+
+studio.initialize()
+const project = getProject('Image type playground', {
+  assets: {
+    baseUrl: 'http://localhost:3000',
```
AriaMinaei, line 14: detect `baseUrl`

theatre/studio/src/panels/DetailPanel/ProjectDetails.tsx
```diff
@@ -34,41 +55,76 @@ const ProjectDetails: React.FC<{
 
   const [downloaded, setDownloaded] = useState(false)
 
-  const exportProject = useCallback(() => {
+  const exportProject = useCallback(async () => {
+    // get all possible asset ids referenced by either static props or keyframes
+    const sheets = Object.values(val(project.pointers.historic.sheetsById))
```
AriaMinaei, line 60: re-use existing fn called `getAllPossibleAssetIDs()`

theatre/studio/src/panels/DetailPanel/ProjectDetails.tsx
```diff
@@ -34,41 +55,76 @@ const ProjectDetails: React.FC<{
...
+        typeof value === 'string' && self.indexOf(value) === index,
+    ) as string[]
+
+    const blobs = new Map<string, Blob>()
+
+    // only export assets that are referenced by the project
+    await Promise.all(
+      allValues.map(async (value) => {
+        const assetUrl = project.assetStorage.getAssetUrl(value)
+        const response = await fetch(assetUrl)
```
AriaMinaei, line 82: `try{...}`

theatre/studio/src/panels/DetailPanel/ProjectDetails.tsx
```diff
@@ -34,41 +55,76 @@ const ProjectDetails: React.FC<{
...
+        if (response.ok) {
+          blobs.set(value, await response.blob())
+        }
+      }),
+    )
+
+    if (blobs.size > 0) {
+      const zip = new JSZip()
+
+      for (const [assetID, blob] of blobs) {
```
AriaMinaei, line 93: Since these are not actually IDs, let's use a different word here. I suggest `assetPath`

theatre/core/src/projects/DefaultAssetStorage.ts
```diff
@@ -0,0 +1,172 @@
...
+
+              // if same same, we do nothing
+              if (sameSame) {
+                console.log('same same', sameSame)
+                return asset.name
+              } else {
+                const renameAsset = (): boolean => {
+                  // if different, we ask the user to pls rename
+                  const newAssetName = prompt(
+                    'Please rename the asset',
```
AriaMinaei, line 137: More descriptive prompt + suggest a new file name

theatre/studio/src/propEditors/simpleEditors/ImagePropEditor.tsx
```diff
@@ -0,0 +1,81 @@
...
+  box-sizing: border-box;
+`
+
+// file input
+const Input = styled.input.attrs({type: 'file', accept: 'image/*'})`
+  display: none;
+`
+
+const Preview = styled.img`
+  position: absolute;
```
AriaMinaei, line 24: The style is hurting my eyes :D 
also: ✋ let me know if you want me to fix it

theatre/studio/src/propEditors/simpleEditors/ImagePropEditor.tsx
```diff
@@ -0,0 +1,81 @@
...
+  cursor: pointer;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  ${({empty}) => empty && `border: 1px dashed currentColor;`}
+  box-sizing: border-box;
+`
+
+// file input
+const Input = styled.input.attrs({type: 'file', accept: 'image/*'})`
```
AriaMinaei, line 19: How do you remove an image?

AriaMinaei (changes requested):
🙌

theatre/core/src/projects/DefaultAssetStorage.ts
```diff
@@ -0,0 +1,172 @@
...
+          },
+        }
+      }
+
+      // get all possible asset ids referenced by either static props or keyframes
+      const possibleAssetIDs = getAllPossibleAssetIDs(project)
+
+      // Clean up assets not referenced by the project. We can only do this at the start because otherwise
+      // we'd break undo/redo.
+      const idbKeys = await idb.keys<string>()
```
AriaMinaei, line 63: Also, let's use a studio-specific key for idb

theatre/core/src/projects/DefaultAssetStorage.ts
AriaMinaei: haha yeah studio _could_ be initialized in ssr actually. only `studio.ui` is guaranteed not to be

theatre/core/src/propTypes/index.ts
AriaMinaei: so `default` should be `{type: 'Image', path: ''}` or something like that

packages/playground/src/shared/image/index.tsx
AndrewPrifer: What do you mean by this?

theatre/core/src/projects/DefaultAssetStorage.ts
AndrewPrifer: Alright, let's exercise an abundance of caution :D

theatre/core/src/projects/DefaultAssetStorage.ts
AndrewPrifer: I found it impossible to filter actual types because these are (afaik) only determined at arbitrary moments during runtime when you call sheet.object(). Filtering for the more general string values is close enough though, and won't break anything.

theatre/core/src/projects/DefaultAssetStorage.ts
AndrewPrifer: Yes

theatre/core/src/projects/DefaultAssetStorage.ts
AndrewPrifer: More descriptive prompt, yes, however I think it is unlikely that any suggestion would be useful. People will just end up doing more work by first deleting the suggestion and then writing their own stuff. I think prefilling the current name is the least annoying option. However if you have something specific in mind, pls let me know.

theatre/core/src/propTypes/index.ts
AndrewPrifer: Yes, we can require default values anyway now that asset ids are user-managed. Although using {type: 'Image', path: ''} instead of a simple string type would make it the first prop type to somehow encode the theatre type in the value instead of the theatre type being determined at runtime through sheet.object() calls so I'm skeptical about that part.

theatre/core/src/propTypes/index.ts
AndrewPrifer: Using {type: 'Image', path: ''} instead of a simple string type would make it the first prop type to somehow encode the theatre type in the value instead of the theatre type being determined at runtime through sheet.object() calls. I'd be curious to hear more about this idea before committing to it.

theatre/studio/src/panels/DetailPanel/ProjectDetails.tsx
AndrewPrifer: I think it is more an ID than a path. xd It is unique like an ID, and while it represents part of the path, in itself it doesn't afford the flexibility that either absolute or relative paths do.

theatre/studio/src/propEditors/simpleEditors/ImagePropEditor.tsx
AndrewPrifer: Well, that's also a discussion for the schema, like how we allow for an empty state, which is also something new as far as prop types go, but for now I added a remove button that just sets the id to ''.

AndrewPrifer:
All the questions open that are worth discussing relate to the schema of the image prop. The rest is minor stuff that I just left open because I partially disagreed with something, but isn't a big deal.

theatre/shared/src/getAllPossibleAssetIDs.ts
```diff
@@ -0,0 +1,22 @@
+import type Project from '@theatre/core/projects/Project'
+import {val} from '@theatre/dataverse'
+
+export default function getAllPossibleAssetIDs(project: Project) {
+  const sheets = Object.values(val(project.pointers.historic.sheetsById))
```
AriaMinaei, line 5: I don't know why TS didn't catch this, but `val(...sheetsById)` could return `undefined`.

theatre/studio/src/panels/DetailPanel/ProjectDetails.tsx
```diff
@@ -57,35 +58,33 @@ const ProjectDetails: React.FC<{
...
-        }
-      }),
-    )
+    try {
+      // only export assets that are referenced by the project
+      await Promise.all(
+        allValues.map(async (value) => {
+          const assetUrl = project.assetStorage.getAssetUrl(value)
+
+          const response = await fetch(assetUrl)
```
AriaMinaei, line 72: this needs a `try{}`

theatre/core/src/propTypes/index.ts
AriaMinaei: Yeah it is a first. We haven't encoded values like this because, so far all value types have been infer-able by their JSON type. The image type is the first one for which we need to encode additional information so it won't be mistaken for another type, like a string, or a hypothetical "AudioFile" type. So I say go for `{type: 'Image', path: string}`.

theatre/core/src/propTypes/index.ts
AriaMinaei: Answered [here](https://github.com/theatre-js/theatre/pull/363#discussion_r1056972282).

theatre/core/src/projects/DefaultAssetStorage.ts
AriaMinaei: Answered [here](https://github.com/theatre-js/theatre/pull/363#discussion_r1056972282)

packages/playground/src/shared/image/index.tsx
AriaMinaei: The `baseUrl` wouldn't work if the host isn't `localhost:3000`

theatre/studio/src/panels/DetailPanel/ProjectDetails.tsx
AriaMinaei: Why not? Can you explain more?

AriaMinaei: But this shouldn't hold the PR from being merged.

AriaMinaei:
I did a code-review, and other than the few issues, this is good to merge.

packages/playground/src/shared/image/index.tsx
AndrewPrifer: Ah, you meant it specifically for the playground. Not sure how we could detect it, or if it was even worth it, since it's just a playground, people can modify it as they please.

theatre/studio/src/panels/DetailPanel/ProjectDetails.tsx
AndrewPrifer: It is not a path because the asset implementation right now, even with the user-managed file names, assumes the assets to be hosted as a flat directory, people cannot specify deeply nested assets. We only identify assets by their file name, which, to me, is just a special case of an ID, not really a path. But I feel like you might be advocating it for calling it a path because you have assumptions about the future implementation of this feature, in which case I'm all ears.:)

theatre/studio/src/panels/DetailPanel/ProjectDetails.tsx
AndrewPrifer: It has one, just around the loop, not the statement itself, so we abort early.

theatre/studio/src/panels/DetailPanel/ProjectDetails.tsx
AndrewPrifer: As for a more general reasoning, I'm calling it ID, because in the end, all assets will have to have a unique id by which they can be identified. The fact that right now that ID is part of their path in the file system seems like an implementation detail that doesn't need to be alluded to in the naming.

PR #255, opened 2022-07-08
Add collapsable items in OutlinePanel

colelawrence:
https://user-images.githubusercontent.com/2925395/178054586-a28da584-38d7-40a7-a9fa-832f1004682a.mp4

Add item collapsing in outline panel

PR #367, opened 2022-12-29
Make items in the outline menu collapsable

clementroche:
Apply requests from #255


clementroche:
Next step is to store collapsed panel data in localStorage

AriaMinaei:
All done. Thank you so much @clementroche 🙌

The only remaining issue is that clicking on the chevron to collapse/exapnd an item, also make that item selected. If you feel like it, send another PR to fix this particular problem. Either way we'll get it fixed before the next release.

mmiinnovations:
> All done. Thank you so much @clementroche 🙌
> 
> The only remaining issue is that clicking on the chevron to collapse/exapnd an item, also make that item selected. If you feel like it, send another PR to fix this particular problem. Either way we'll get it fixed before the next release.

Overall I want to be able to use TheatreJs to animate/control every prop I want, including DOM (CSS, div props, React state changes, etc.). Controlling and animation of opacity is a major use case that's giving me fits currently.

Example, I want to be able to use TheatreJs to animate HTML form after user inputs data. If form returns data I want animate 3D objects (like a 3D toast or error div) over the 2D DOM UI using Theatre sheets timeline animations.

PR #368, opened 2022-12-30
feat(step): ticket 314

tomorrowevening:
Step functionality https://github.com/theatre-js/theatre/issues/314

### DIFF
--- a/theatre/core/src/propTypes/index.ts
+++ b/theatre/core/src/propTypes/index.ts
@@ -254,6 +254,7 @@ export const number = (
     nudgeFn?: PropTypeConfig_Number['nudgeFn']
     range?: PropTypeConfig_Number['range']
     nudgeMultiplier?: number
+    step?: number
     label?: string
   } = {},
 ): PropTypeConfig_Number => {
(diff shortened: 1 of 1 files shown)


AriaMinaei:
Ok I just pushed fd3a8cec35bbcf00dff12ff6534da24f8376b960 which fixes #314 itself. It simply fixes the behavior of `nudgeMultiplier` so it acts like `step` in this PR. However, this PR goes \*ahem\* a _step_ further and makes `step` act as a type constraint, rather than just a UX affordance. Here is what I mean:

If the user sets `step: 10`, and then sets two keyframes, from `0` to `100`, then the easing between those keyframes will be `0, 10, 20, 30, ..., 100`. IOW, the easing wouldn't be smooth. So, basically, `step=10` will make `number` behave like an integer with increments of `10`, while `nudgeMultiplier=1` will keep `number` act as a `float`, but make its nudging be in increments of `1`.

Is that the behavior you are going for @tomorrowevening? And how will you use it? What kind of props will have `step` rather than just `nudgeMultiplier`?

clementroche:
In my use case i have a `count` prop that control length of an array so I want it to be integer and have 0 as minimum in any circunstance

PR #374, opened 2023-01-05
Feature: Custom RAFDrivers

### LINKED ISSUE #39: Ability to replace the scheduler for a custom one

AriaMinaei:
This PR allows you to control when computations in Theatre tick forward.

We call this mechanism a `rafDriver` (raf stands for [`requestAnimationFrame`](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame)).

The default `rafDriver` in Theatre creates a `raf` loop and ticks forward on each frame. With this PR, you can create your own `rafDriver`, which enables the following use-cases:

1. When using Theatre.js alongside other animation libs (`@react-three/fiber`/`gsap`/`lenis`/`etc`), you'd want all animation libs to use a single `raf` loop to keep the libraries in sync and also to get better performance.
2. In XR sessions, you'd want Theatre to tick forward using [`xr.requestAnimationFrame()`](https://developer.mozilla.org/en-US/docs/Web/API/XRSession/requestAnimationFrame).
3. In some advanced cases, you'd just want to manually tick forward (many ticks per frame, or skipping many frames, etc). This is useful for recording an animation, rendering to a file, testing an animation, running benchmarks, etc.

This PR resolves #39, replaces #230, and #88.

Here is how you'd create a custom `rafDriver`:

```js
import {createRafDriver} from '@theatre/core'

const rafDriver = createRafDriver({name: 'a custom 5fps raf driver'})

setInterval(() => {
  rafDriver.tick(performance.now())
}, 200)
```

Now, any time you set up an `onChange()` listener, pass your custom `rafDriver`:

```js
import {onChange} from '@theatre/core'

onChange(
  // let's say object is a Theatre object, the one returned from calling `sheet.object()`
  object.props,
  // this callback will now only be called at 5fps (and won't be called if there are no new values)
  // even if `sequence.play()` updates `object.props` at 60fps, this listener is called a maximum of 5fps
  (propValues) => {console.log(propValues)},
  rafDriver
)

// this will update the values of `object.props` at 60fps, but the listener above will still get called a maximum of 5fps
sheet.sequence.play()

// we can also customize at what resolution the sequence's playhead moves forward
sheet.sequence.play({rafDriver}) // the playhead will move forward at 5fps
```

You can optionally make studio use this `rafDriver`. This means the parts of the studio that tick based on raf, will now tick at 5fps. This is only useful if you're doing something crazy like running the studio (and not the core) in an XR frame.

```js
studio.initialize({
  __experimental_rafDriver: rafDriver,
})
```

`rafDriver`s can optionally provide a `start/stop` callback. Theatre will call `start()` when it actually has computations scheduled, and will call `stop` if there is nothing to update after a few ticks:

```js
import {createRafDriver} from '@theatre/core'
import type {IRafDriver} from '@theare/core'

function createBasicRafDriver(): IRafDriver {
  let rafId: number | null = null
  const start = (): void => {
    if (typeof window !== 'undefined') {
      const onAnimationFrame = (t: number) => {
        driver.tick(t)
        rafId = window.requestAnimationFrame(onAnimationFrame)
      }
      rafId = window.requestAnimationFrame(onAnimationFrame)
    } else {
      driver.tick(0)
      setTimeout(() => driver.tick(1), 0)
    }
  }

  const stop = (): void => {
    if (typeof window !== 'undefined') {
      if (rafId !== null) {
        window.cancelAnimationFrame(rafId)
      }
    } else {
      // nothing to do in SSR
    }
  }

  const driver = createRafDriver({name: 'DefaultCoreRafDriver', start, stop})

  return driver
}
```

A couple of design decisions:

1. You can have several `rafDriver`-s on the same page, or even listening to the changes in values of the same object. 
  
  ```js
  onChange(object.props, listener, aFiveFPSDriver)
  onChange(object.props, listener, a60FPSDriver)
  ```
  
2. I debated whether to allow setting a default `rafDriver` for a whole project or all of `@theatre/core`. That's certainly possible, and it would make the API less verbose. However, it also makes the API more ambiguous (like which driver takes precedence when there are several?) and more tricky (you'd have to call the hypothetical `core.setDefaultRafDriver()` before any computation is done).

3. We're not exposing `Ticker` because a `Ticker` has responsibilities beyond just scheduling, such as keeping a list of stale computations and (in the future) prioritizing them.

### DIFF
--- a/packages/benchmarks/src/index.tsx
+++ b/packages/benchmarks/src/index.tsx
@@ -1,5 +1,5 @@
 // import studio from '@theatre/studio'
-import {getProject} from '@theatre/core'
+import {createRafDriver, getProject} from '@theatre/core'
 import type {
   UnknownShorthandCompoundProps,
   ISheet,
@@ -7,11 +7,11 @@ import type {
 } from '@theatre/core'
 // @ts-ignore
 import benchProject1State from './Bench project 1.theatre-project-state.json'
-import {Ticker} from '@theatre/dataverse'
-import {setCoreTicker} from '@theatre/core/coreTicker'
+import {setCoreRafDriver} from '@theatre/core/coreTicker'
 
-const ticker = new Ticker()
-setCoreTicker(ticker)
+const driver = createRafDriver({name: 'BenchmarkRafDriver'})
+
+setCoreRafDriver(driver)
 
 // studio.initialize({})
 
@@ -79,7 +79,7 @@ async function test1() {
   }
 
   function iterateOnSequence() {
-    ticker.tick()
+    driver.tick(performance.now())
     const startTime = performance.now()
     for (let i = 1; i < CONFIG.numberOfIterations; i++) {
       onChangeEventsFired = 0
@@ -87,7 +87,7 @@ async function test1() {
       for (const sheet of sheets) {
         sheet.sequence.position = pos
       }
-      ticker.tick()
+      driver.tick(performance.now())
       if (onChangeEventsFired !== objects.length) {
         console.info(
           `Expected ${objects.length} onChange events, got ${onChangeEventsFired}`,
--- a/packages/dataverse/src/Ticker.ts
+++ b/packages/dataverse/src/Ticker.ts
@@ -1,6 +1,14 @@
 type ICallback = (t: number) => void
 
 /**
+ * The number of ticks that can pass without any scheduled callbacks before the Ticker goes dormant. This is to prevent
+ * the Ticker from staying active forever, even if there are no scheduled callbacks.
+ *
+ * Perhaps counting ticks vs. time is not the best way to do this. But it's a start.
+ */
+const EMPTY_TICKS_BEFORE_GOING_DORMANT = 60 /*fps*/ * 3 /*seconds*/ // on a 60fps screen, 3 seconds should pass before the ticker goes dormant
+
+/**
  * The Ticker class helps schedule callbacks. Scheduled callbacks are executed per tick. Ticks can be triggered by an
  * external scheduling strategy, e.g. a raf.
  */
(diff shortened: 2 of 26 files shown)


AriaMinaei:
What do you think @felthy and @donmccurdy? Does this address all of the requirements you mentioned?

A build of this PR is on npm on version `0.6.0-insiders.954391e`.

felthy:
Thank you for pursuing this @AriaMinaei! This API looks like it will meet our needs for all of our use cases.

I tested this today, and it works nicely when I want to manually advance the playhead and flush changes, by setting `sequence.position` then calling `tick()` on my driver.

However if I use `sequence.play()`, the internal `coreTicker` is being relied upon to advance the playhead.
* There is no public API to allow me to use another driver, like `XRSession`, for this.
* If there was, all of my sheets would be driven by the same driver. I need to isolate each driver to its sheet, such that calling `tick()` on one driver is guaranteed not to affect any other sheets.
* Even if it will do so by default for convenience, I should be able to prevent Theatre from ever calling `requestAnimationFrame()` internally.

From what I can tell, the only things using the `coreTicker` are the `PlaybackController` implementations. I notice they only use the ticker from inside their `play()` and `playDynamicRange()` methods, so it seems we could:
* remove the `ticker` constructor param (and private property) from the `PlaybackController` implementations
* have the public `sequence.play()` and `sequence.playDynamicRange()` methods accept an optional `rafDriver`, and pass it through to the `PlaybackController` methods

Then if I always provide my custom `rafDriver` when calling `play()`, a `coreTicker` will never be created and I'll have no need to explicitly disable it.

What do you think?

AriaMinaei:
Thanks for testing this out, @felthy! I'll make `sequence.play()` take an optional `driver` param. 

Beside that, I'm curious, why would you want to "prevent Theatre from ever calling `requestAnimationFrame()` internally"?

AriaMinaei:
@felthy okay, with this last commit, calling `sequence.play({rafDriver})` is now possible. But the core `requestAnimationFrame()` driver will keep ticking as well, even if its Ticker's queue is empty. I'll do that next.

AriaMinaei:
Alright, [a24a77c](https://github.com/theatre-js/theatre/pull/374/commits/a24a77c8b2422f86d7432c9f8406cb09b1cb108e) implements active/dormant tickers. This means that if every `onChange()` and `sequence.play()` call uses a custom `rafDriver`, then the default driver that calls `requeseAnimationFrame()` won't even start.

clementroche:
Can we add rafDriver to a onValuesChange callback ?

PR #376, opened 2023-01-10
change component length limit from 32 characters to 64

AD2018:


AriaMinaei:
Merged. Thanks @AD2018!

PR #393, opened 2023-02-03
Experimental API for forgetting objects/sheet

AriaMinaei:
This adds two methods to the `transactionAPI` that would remove an object or sheet's state from the project's state.

```ts
import {getProject} from '@theatre/core'
import studio from '@theare/studio'

const project = getProject('...')
const sheet = project.sheet('...')
const object = sheet.object('...')

// let's say you've already set up prop values for your object and created a sequence in the studio

studio.transaction((api) => {
  // calling this will make it as if you never set values for this object or put it in a sequence
  api.__experimental_forgetObject(object)
  
  // calling this will make it as if you never set values for _any_ object in this sheet, and you never created a sequence either.
  api.__experimental_forgetSheet(sheet)

  // note that if you're calling __experimental_forgetSheet(), then there is no need to call __experimental_forgetObject() in case that object belongs in that sheet.
})

// now if you export the project's state, or do that programmatically, the sheet/object won't be in the exported state:
studio.
createContentOfSaveFile(project.address.projectId) // won't have the sheet/object in it

// note that cmd/ctrl+z will still undo the above operations
```

### DIFF
--- a/theatre/studio/src/TheatreStudio.ts
+++ b/theatre/studio/src/TheatreStudio.ts
@@ -19,6 +19,8 @@ import {
   __experimental_disblePlayPauseKeyboardShortcut,
   __experimental_enablePlayPauseKeyboardShortcut,
 } from './UIRoot/useKeyboardShortcuts'
+import type TheatreSheetObject from '@theatre/core/sheetObjects/TheatreSheetObject'
+import type TheatreSheet from '@theatre/core/sheets/TheatreSheet'
 
 export interface ITransactionAPI {
   /**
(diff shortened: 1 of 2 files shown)


AriaMinaei:
Published an insiders' release as `0.6.1-insiders.7dde8ab`

AriaMinaei:
This might interest @tomorrowevening too.

tomorrowevening:
I'll experiment with this later with my drag and drop prototype, that dynamically adds sheets / objects and would be a good test for this

AD2018:
Please add some method(`detachSheet()`) that will hide sheet from outline menu in the top left so people can also hide it from UI after deleting sheet

AriaMinaei:
@AD2018 sure, will do

f0nar:
@AriaMinaei do you have any updates for removing of sheets from project tree?

AriaMinaei:
@f0nar This was [merged](https://www.theatrejs.com/docs/latest/releases#v0.6.1) in 0.6.1. Upgrade to that version (well, better latest, so 0.7.0) and you'll have the API described in the [top comment](https://github.com/theatre-js/theatre/pull/393#issue-1570315016).

f0nar:
@AriaMinaei I was asking about functionality like 'detachSheet'

AriaMinaei:
@f0nar  `detachSheet()` hasn't been implemented yet, but `detachObject()` is: https://www.theatrejs.com/docs/latest/manual/objects#detaching-objects

Do you specifically need to detach sheets from a project?

f0nar:
Yep. I really need it.

PR #396, opened 2023-02-06
optimize builds

AriaMinaei:
We used to let the user's bundler config determine if the studio should be optimized for production. This only happened if the user set `process.env.NODE_ENV === "production"` and turned on tree-shaking.

Since most users run studio in a dev server, they won't see any benefits from these optimizations. This change enables these optimizations for `@theatre/studio`. However, `@theatre/core` still depends on the user's bundler config.


AriaMinaei:
Rebased, so I'll make a new release

PR #408, opened 2023-03-13
"File Prop" Type

joaogsleite:
There is already an "Image" Prop Type but it is not possible to select other file types besides images. 

With this new "File" Prop Type, it will be possible to select any file type.

### Example use case:
Select *glb* file to change model.

### DIFF
--- a/packages/playground/src/shared/file/index.html
+++ b/packages/playground/src/shared/file/index.html
@@ -1,0 +1,11 @@
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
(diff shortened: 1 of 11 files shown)


AriaMinaei:
This is a great PR. Thank you @joaogsleite! I'll review this week.

akre54:
This looks great! In addition to 3d models we'd love to use this for importing data like CSVs or JSONs. Would love to see this merged!

AriaMinaei:
Merged. Thank you @joaogsleite :)

PR #409, opened 2023-03-23
feat(dataViewer): component introduced

tomorrowevening:
A sheet-specific way to visualize data during animating.
This would require a way to attach data to sheets - this data would not be exported out with the main JSON as it's there only as a helper during development.


tomorrowevening:
I need some help resizing the component when the timeline's bottom scroller expands/decreases

PR #417, opened 2023-05-04
fix(CurveEditor): error when copy/pasting eases

tomorrowevening:
Fixes the bug when a user copies the easing preset and tries pasting it somewhere else:

`Uncaught TypeError: Cannot read properties of undefined (reading 'target') at onInputChange (CurveEditorPopover.tsx:178:21)`


AriaMinaei:
Thanks @tomorrowevening!

PR #422, opened 2023-05-30
Feat/sharable sequence data

tomorrowevening:
Import / export sequence data

- [x] Markers
- [x] Custom Data


tomorrowevening:
@AriaMinaei should we still try and get this in?

Issue #419, opened 2023-05-19
Interpolate Track values on curve


akre54:
Hi there - I'd like to get all values for a given Track to be able to add a "trail" effect without statefully adding to a list of "seen" values.

Currently the `types` support a linear interpolator that maps between a beginning and end value, but the result doesn't align with the bezier curve of the timeline's actual keyframes. I'd like to be able to be able to [retrieve a value for a given time](https://github.com/theatre-js/theatre/issues/330#issuecomment-1411080554) or a quantized list of values for the whole sequence time, similar to [D3's interpolateBasis](https://observablehq.com/@d3/d3-interpolatebasis). Currently it requires a fair bit of hacking in userland. Is this something that could be added to the library?

AriaMinaei:
This sounds like a great feature to have. Maybe `seq.getKeyframes(object.props.x) => Array<Keyframe>` or something like that. It's not yet on the roadmap but I'll be happy to merge a PR if one shows up ;)

akre54:
Yes I can definitely look into this - finding it a bit tricky to wade through that part of the code but I'll try to whip something up.

akre54:
Created: #426

PR #426, opened 2023-06-12
Add a getKeyframes method to Sequence

akre54:
Adds the ability to get the keyframes for a given SheetObject prop. Please feel free to tear it apart or modify as you see fit.

Ref https://github.com/theatre-js/theatre/issues/419#issuecomment-1565390407


AriaMinaei:
Alright, this one is in. Thank you @akre54! I added a test, and renamed the method to `__experimental_getKeyframes()` in case we'd need to change its interface later.

PR #430, opened 2023-07-05
feat(remote): multiple-window support

tomorrowevening:
Functionality to control theatre from another window.

### DIFF
--- a/packages/playground/src/shared/remote/Box3D.tsx
+++ b/packages/playground/src/shared/remote/Box3D.tsx
@@ -1,0 +1,85 @@
+import React, {useEffect, useRef} from 'react'
+import type {CSSProperties} from 'react'
+import {types} from '@theatre/core'
+import type {ISheet} from '@theatre/core'
+import {remote} from './Remote'
+
+// Box element
+export const BoxSize = 100
+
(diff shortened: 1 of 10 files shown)


AriaMinaei:
Thanks @tomorrowevening! Btw, if you get around to it, a small readme for this playgorund would be helpful.

vimlesh1975:
How to use it?

tomorrowevening:
This is just a proof of concept so feel free to make it your own.
[Remote](https://github.com/theatre-js/theatre/blob/main/packages/playground/src/shared/remote/Remote.ts) works as a Singleton that contains all animation & configuration that is shared by both your application and the editor, with [RemoteController](https://github.com/theatre-js/theatre/blob/main/packages/playground/src/shared/remote/RemoteController.ts) as the mediator that only gets initiated during development mode.
When running `yarn playground`:
Open 1 window to [http://localhost:8082/shared/remote](http://localhost:8082/shared/remote)
And another window to access the editor mode:  [http://localhost:8082/shared/remote#editor](http://localhost:8082/shared/remote#editor)

vimlesh1975:
Thank you. I tested it. Good feature.

PR #433, opened 2023-07-12
fix(geometry): PlaneBufferGeometry deprecated, used PlaneGeometry

tomorrowevening:
A lot of ppl are complaining they have to use an older version of ThreeJS because of this (<v154)


AriaMinaei:
Thank you @tomorrowevening!

Issue #411, opened 2023-03-29
Setting keyframes from code


hybridherbst:
I'm creating some procedural animations, and would like to visualize them and continue editing with theatre.
I looked into some of the code for the detail pane, and things like setKeyframeAtPosition, however I can't get that to work / it doesn't seem to be documented.

Is there an example for something like "create 10 keyframes for a specific object in code and continue to work on it in theatre"?

Thanks!

vimlesh1975:
I am a beginner in this theatrejs.
I have tested and found something like below works.
`  studio.transaction((api) => {
            for (let i = 0; i <= 10; i++) {
                sheet.sequence.position = i;
                api.set(obj1.props.left, i * 100);
            }
        })`

akre54:
Late to this, but we're currently running up against this same issue. It'd be nice if there were a little cleaner API but it's definitely doable.

Automating the track data would require calling something like stateEditors.ts `_ensureTracksOfObject` from within `set`. Is this something you'd be open to @AriaMinaei ?

PR #449, opened 2023-09-16
Add __experimental_sequenceProp to studio.transaction()

akre54:
I want the ability to programmatically add keyframes to a sequence, but it looks like the best way to do this is to use `transaction(({set}) => {})`. Unfortunately `set` doesn't create a Track from an unsequenced prop.

So let's say I have some data I want to turn into keyframes.

```ts
studio.transaction(({ set, unset }) => {
  const pos = sequence.position
  for (const keyframe of sequence.__experimental_getKeyframes(obj.props.x)) {
    sequence.position = keyframe.position
    unset(obj.props.x)
  }
  for (const {time, value} of data) {
    sequence.position = time
    set(obj.props.x, value)
  }
  sequence.position = pos
})
```

Currently, on `main`, If `obj.props.x` isn't already sequenced (i.e. manually from the Studio UI) then `set` will just set the static prop, which isn't what we want here. This pull allows you to at least tell Studio that you want to sequence the property.

@vimlesh1975 [has some code](https://discord.com/channels/870988717190426644/870988717190426647/1152277162305519687) that he [posted to Discord](https://discord.com/channels/870988717190426644/870988717190426647/1152277162305519687) that loops over all keyframes and unsets their value at the given time, but it would be really nice to be able to clear all keyframes for the initial track data, and ensure that the track is sequenced ahead of time. 

See also #411 

Ideally I'd like something like:
```ts
sequence.clearKeyframes(obj.props.x)
sequence.setKeyframeAtTime(obj.props.x, time, val)
```

But in the meantime at least this fixes the ambiguity of whether you want to set a static prop or a sequenced prop.


akre54:
Changed to draft. `getArrayOfValidSequenceTracks` (called by `set`) returns empty, so this needs to be fixed first

AriaMinaei:
Hi @akre54, this one is merged via 74b68c8f12698bd4b1c46c14f550295625630f88. Feel free to send PRs for these as well 🙏

```ts
sequence.clearKeyframes(obj.props.x)
sequence.setKeyframeAtTime(obj.props.x, time, val)
```

akre54:
Hey @AriaMinaei any hope for this to be released soon? I have a feature that deals a lot with editing the timeline. Thanks! 🙏

Issue #514, opened 2025-11-17
False conflict detection when revisionHistory is empty


White-Devil2839:
False conflict detection when revisionHistory is empty – packages/core/src/projects/initialiseProjectState.ts looks for the most recent on‑disk revision and marks the project as conflicted whenever it can’t find that revision in the browser state. When the shipped conf.state has never been exported with a revision (its revisionHistory is [], which is the default produced inside the repo), the lookup uses undefined and always falls into the “conflict” branch even if there’s no real divergence. This locks new projects into the red “Browser state is not based on disk state” warning until the user manually resolves it every reload, despite both states being compatible.

Issue #509, opened 2025-08-11
sheet.detachObject doesn't clear staticOverrides


akre54:
If I create an object with `sheet.object('obj', { num: types.number(0) })`, set the value of `num` to 20 in the UI, and then later call [`sheet.detachObject('num')`](https://github.com/theatre-js/theatre/blob/0.7.1/theatre/core/src/sheets/TheatreSheet.ts#L256), I'd expect that the exported state would be empty. Instead I have a key in `staticOverrides` for the prop.

I can verify this by calling `sheet.object` again with the same config, and the UI shows its value as 20. My expectation would be that the number should reset to the default 0 value, and that I wouldn't get an exported prop.

It looks like the internal Sheet class [only deletes the object](https://github.com/theatre-js/theatre/blob/0.7.1/theatre/core/src/sheets/Sheet.ts#L81-L87) from `this._objects` and not any of the dataverse pointers.

Is this an easy fix?

White-Devil2839:
I have resolved it please assign this issue to me @akre54

akre54:
I don't think I can assign the issue to you... Only @AriaMinaei can

White-Devil2839:
@AriaMinaei can you assign this issue to me please as I have already solved this and also thank you very much @akre54 for helping me

akre54:
Do you have a fork of the repo up with the branch? That might be the best way to share changes

White-Devil2839:
Yes @akre54

akre54:
Can you push it up to github and paste a link? That way it'll show up in the [network graph](https://github.com/theatre-js/theatre/network)

White-Devil2839:
Oh Thank you very much I didn't think of that @akre54 I will make sure to do it

White-Devil2839:
hi @akre54 https://github.com/theatre-js/theatre/pull/517
I created the pull request and resolved the issues but it shows deployment issue

PR #516, opened 2025-11-17
Fix sheet detach to clear overrides

White-Devil2839:
I have fixed the created issue #514 and #509


White-Devil2839:
@AriaMinaei can you help

PR #517, opened 2025-11-18
I have fixed issues #514 and #509

White-Devil2839:
Issues Resolved

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 
 ## 0.4.3
 
+
 * New features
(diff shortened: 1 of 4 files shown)


White-Devil2839:
@akre54 Can you help me here it's working on my local and I updated node to v22 already but it shows same deployment error

akre54:
It seems to be an issue on the vercel config side. @AriaMinaei would have to make the change. 

But this is helpful! Thanks for making the change

White-Devil2839:
Thank you for the feedback @akre54 can u help me get it accepted by @AriaMinaei

Issue #491, opened 2024-05-15
Shared/dom route on playground website is not working


deepansh96:
![image](https://github.com/theatre-js/theatre/assets/20169829/99685612-277e-4ce7-a3c6-43d5a8440f1c)

This website - `https://theatre-playground.vercel.app/shared/dom` mentioned in the docs doesn't open up and throws the error shown above

PR #521, opened 2025-12-03
fix crash in shared/dom route (#491)

jagadish-555:
This PR fixes a crash where the DOM playground scene wouldn’t load due to a Studio initialization race condition.

What’s changed:

In Scene.tsx, replaced getStudioSync() with an async getStudio().then() call so we wait for Studio to initialize before using it — preventing the “undefined” access crash.

In index.tsx, added import '@theatre/studio' to ensure the Studio bundle is included in the build, fixing the 404s and the “Project state is empty” issue.


Alchemy86:
This still an ongoing project?

